PR #12, opened 2015-01-08
Don't generate sourcemaps if user disables them

gaearon:
Currently, source maps are always generated, even if Webpack doesn't use them.

However Webpack provides [`this.sourceMap`](http://webpack.github.io/docs/loaders.html#sourcemap) to loaders.
This flag will only be `true` if user enabled source maps in their Webpack config.

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,13 +21,16 @@ module.exports = function (source) {
         return accumulator;
     }, {});
 
-    options.sourceMap = true;
+    options.sourceMap = this.sourceMap;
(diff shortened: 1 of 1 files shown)


sebmck:
Thanks! Released as of 2.0.0

gaearon:
It's not on NPM yet, is it?

sebmck:
@gaearon It was available on npm as soon as I merged.

gaearon:
Whoops, I installed 0.2.0 by mistake.
Thanks for reacting quick!

PR #13, opened 2015-01-09
Update README.md

nhunzaker:
I had to make this change in order to get the runtime to inject properly. Did I miss something?

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ plugins: [
   // Alternatively, write `require('6to5/runtime')` at the top of your entry point.
   // Leaks the object to the window, but it's simple.
   new webpack.ProvidePlugin({
-    to5Runtime: "imports?global=>{}!exports-loader?global.to5Runtime!6to5/runtime"
+    to5Runtime: "6to5/runtime"
   })
 ]
 ```


sebmck:
Ping @Couto

Couto:
I did a quick test with a current project and your configuration always results in errors.
Just to make sure, did you enable the runtime option?

nhunzaker:
I believe so. This is how I have the loader configured at any rate:

```javascript
{
  test    : /\.jsx*$/,
  exclude : /node_modules/,
  loader  : 'envify-loader!6to5-loader?experimental=true&modules=common'
},
```

nhunzaker:
Aye, and here's the error I'm getting:

```
Module not found: Error: Cannot resolve module 'imports' in {file}
```

To produce this error, I have the loader configured like so:

```javascript

module.exports = {
  plugins: [
    new WebPack.ProvidePlugin({
      to5Runtime: "imports?global=>{}!exports-loader?global.to5Runtime!6to5/runtime"
    })
  ],
  module: {
    loaders: [
      {
        test    : /\.jsx*$/,
        exclude : /node_modules/,
        loader  : 'envify-loader!6to5-loader?experimental=true&modules=common&runtime=true'
      }
    ]
  }
}

**Comment by Couto** [comment]
You need to install the loaders: `npm install imports-loader exports-loader --save-dev` since they aren't included by default with webpack.

**Comment by nhunzaker** [comment]
Ahhhh, of course. Thoughts on including this in the readme?

**Comment by Couto** [comment]
I will.
Could you confirm me that this doesn't work?

```
module.exports = {
  plugins: [
    new WebPack.ProvidePlugin({
      to5Runtime: "6to5/runtime"
    })
  ],
  module: {
    loaders: [
      {
        test    : /\.jsx*$/,
        exclude : /node_modules/,
        loader  : '6to5-loader?experimental&runtime&modules=common'
      }
    ]
  }
}
```
Because it would really be nice that the example above could in fact work, but I think it doesn't

nhunzaker:
It _does_ work!

nhunzaker:
Well.. actually. It just successfully compiles. It looks like the run time is not present. Sorry for jumping the gun

nhunzaker:
@Couto curiously, if I remove the `runtime` flag, your example does work. Any idea why this is the case?

Dashed:
@Couto Probably need to clarify that `imports-loader exports-loader` are needed as deps, especially for folks not aware of webpack loaders.

nhunzaker:
Yeah, I think that's the key issue. I think even a quick "install imports-loader and exports-loader" would be sufficient

Couto:
I've just added the instructions to install the loaders. Since English is not my first language I would appreciate some revision.

nhunzaker:
Looks good to me!

PR #17, opened 2015-01-20
README: Use ES6 syntax highlighting

Turbo87:


Dashed:
Should eventually be using `javascript` in the future.

thejameskyle:
You can use `js` or `javascript` now, they're highlighted exactly the same way.

Turbo87:
I've updated the PR to use `javascript` instead of `es6`. Thanks for the suggestion!

PR #22, opened 2015-01-29
typo left over from 2.x

jquense:


sebmck:
The `experimental` option actually just enables [experimental mode](6to5.org/docs/usage/experimental). If that part can be removed from the comment then this should be good to merge!

jquense:
sorry for the extra commit, doing via the github editor :P

sebmck:
All good, thank you!

PR #30, opened 2015-02-15
Change 6t05 to babel in the package.json description

hawkrives:
Should I also add the `babel` keyword?

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "webpack",
     "loader",
     "6to5",
+    "babel",
     "es6",
     "transpiler",
     "module"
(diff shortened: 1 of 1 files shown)


Couto:
Yes. I believe that adding the keyword babel would improve search results. 
Would you mind adding it to the PR?

hawkrives:
Done! 

I think that we should probably leave the 6to5 keyword, at least for now, so that it still shows up for people searching for 6to5.

Couto:
I agree.
Thanks for the PR :)

hawkrives:
No prob. Thanks for making things!

PR #33, opened 2015-02-24
Update webpack and babel version

jtangelder:


sebmck:
They're caret versions so this isn't necessary. Whenever you install `babel-loader` you'll get the latest version.

Issue #34, opened 2015-02-26
Feature Query: Filesystem Cache?


loganfsmyth:
I work on a pretty big app and transpiling everything take a while on the first webpack build pass. Would you be open to a PR to add a filesystem cache to this module like `babel/register` does? I implemented it in a custom loader and for us, it reduces the initial build pass by ~20 seconds. I'd be happy to submit a PR to add it to the this as a optional flag.

Couto:
Any way to see your custom loader? Just to check the inner workings. 
But as far as I can tell that sounds like an excellent PR :+1:

loganfsmyth:
@Couto This is what I had thrown together for my test: https://gist.github.com/loganfsmyth/b0c0ed5d9f6eafecd5cc

It's a few weeks old (still using "6to5") and has some purpose-specific stuff for our usage, but it demonstrates what I mean. We're not acutally using this yet, I just have it on a branch. I'd want to change to using `path.join(os.tmpdir(), '/babel-cache')` or something for the dir. I have it caching based on content, babel version, and args, which I think makes the most sense. Probably want the cache location to be configurable by param too.

It takes out build times from 39 -> 22 seconds once the cache is populated, so restarting our watcher scripts and stuff is much faster and it seems like something that others would benefit from, at least as an optional thing.

Couto:
I have to admit that I didn't have the opportunity to test it, but as far as I can see, I'm pretty ok with a PR, specially if it is behind a flag :)

On the other hand, I feel that you could also direct this PR directly to webpack so that all loaders could benefit of the cache improvements, but it's just an idea.

jtangelder:
This would be cool, our build takes about 2 minutes initial, which is way too long.

loganfsmyth:
I'm going to try to get a PR together soon.

loganfsmyth:
@jtangelder Submitted a PR, I'd be interested to hear how that works for you.

jtangelder:
Awesome, initial build now is at 50% of the time!

Couto:
You guys can find this on version 4.2.0.

I'll try to ring the guys at webpack at gitter to see if they want to consider something similar to other loaders.

jtangelder:
A webpack 'cache-loader' would make sense :-)

Couto:
@jtangelder I agree actually. I've merged because the benefits are indeed great, but I also believe that this shouldn't live specifically inside the babel-loader, and that other loaders should benefit, either by updating webpack or an independent loader :+1:

PR #41, opened 2015-03-13
Support an optional 'cacheDirectory' param to cache babel loader results.

loganfsmyth:
Here's a first pass at my solution for #34. Added an explicit cache param and `zlib` usage to keep the size of the cache a bit more reasonable.

Our codebase is pretty big, using this cache, I see:

* Transpiling both source and tests: `130s` with cold cache, `40s` with cache, 3.25x speedup
* Transpiling only source `45s` with cold cache, `22s` with cache, 2x speedup

Those numbers definitely make it clear to me that having the cache will be a big help to webpack initial start time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,13 @@ loaders: [
 
 See the `babel` [options](http://babeljs.io/docs/usage/options/)
 
+This loader also supports the following loader-specific option:
+
+* `cacheDirectory`: When set, the given directory will be used to cache the results of the loader.
+  Future webpack builds will attempt to read from the cache to avoid needing to run the potentially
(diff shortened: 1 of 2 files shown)


Couto:
Thanks a lot for this! :)

Couto:
[cache refactor](https://github.com/babel/babel-loader/blob/feature/refactor-cache/lib/fs-cache.js)
@loganfsmyth Would you mind to take a look at the link? 

I think this code could easily live in it's own package (something like 'fs-cache-loader-util' or something) 
Since this is mostly your code, do you want to publish it in a self contained package? If you don't, would you mind that I would? 

Thanks :)

PR #42, opened 2015-03-22
Add note to readme to clarify sourceMap behaivor

Rob-ot:


graingert:
@Couto @Rob-ot are you sure this is the case, it looks like it's being pulled in if it's not undefind.

graingert:
https://github.com/babel/babel-loader/blob/master/index.js#L55

PR #57, opened 2015-05-01
Move babel-core to devDependencies

kossnocorp:
Prevent `TypeError: Transformer PLUGIN NAME is resolving to a different Babel version to what is doing the actual transformation…` errors by removing `babel-core` from `dependencies`.

See: https://github.com/babel/babel/issues/1244


kossnocorp:
Whoops, I messed up this PR.

PR #58, opened 2015-05-01
Remove babel-core from dependencies

kossnocorp:
Prevent `TypeError: Transformer PLUGIN NAME is resolving to a different Babel version to what is doing the actual transformation…` errors by removing `babel-core` from `dependencies`.

See: https://github.com/babel/babel/issues/1244

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,8 +1,22 @@
 var assign = require('object-assign');
-var babel = require('babel-core');
 var cache = require('./lib/fs-cache.js');
 var loaderUtils = require('loader-utils');
 var pkg = require('./package.json');
+
+try {
+  var babel = require('babel-core');
+
+} catch(err) {
+  if (err.code != 'MODULE_NOT_FOUND') {
+    throw err;
+  }
+
+  console.error(
+    'Error: babel-core package is not installed, please run:\n\n' +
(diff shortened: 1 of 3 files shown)


sebmck:
Why? This is just going to make it more difficult to use `babel-loader`.

kossnocorp:
@sebmck this is trade-off, having `babel-core` in dependencies will cause even more problems when user will start using plugins.

sebmck:
@kossnocorp Not really. In 5.2.0 there's an alternate plugin API. See [babel/babel#1244](https://github.com/babel/babel/issues/1244#issuecomment-97991260).

index.js
```diff
@@ -1,9 +1,23 @@
...
+  if (err.code != 'MODULE_NOT_FOUND') {
+    throw err;
+  }
+
+  console.error(
+    'Error: babel-core package is not installed, please run:\n\n' +
+    '    npm install --save-dev babel-core\n'
+  );
+  process.exit(1);
+}
```
kossnocorp: @sebmck in case when user don't have `babel-core` in dependencies we will render this nice error message.

kossnocorp:
@sebmck if so, feel free to close this PR.

PR #66, opened 2015-05-16
Fix reading cache function to callback a proper result

deepsweet:
hi.

to solve `Module build failed: Error: Final loader didn't return a Buffer or String` which is constantly appears after `@5.1` update.

`content` is a raw file buffer, so `.code` from it is `undefined`.


Couto:
I've cherry-picked this commit so it's already in the master and on npm with version 5.1.2 I'm going to close it :)

Thanks for the PR :)

PR #70, opened 2015-05-20
Default value for cacheDirectory

gpbl:
If set to `true`, it may cause the error described below in the doc... i changed the readme to prevent confusion.


Couto:
:+1:

PR #60, opened 2015-05-07
support sync compilation

camshaft:
this is needed for using the [enhanced-require](https://www.npmjs.com/package/enhanced-require)

PR #79, opened 2015-06-14
Enter async mode only when necessary

jscheid:
Hi,

Ever since babel-loader introduced caching we can't use it anymore together with [enhanced-require](https://github.com/webpack/enhanced-require) as that doesn't support async callbacks in loaders.

This change restores enhanced-require compatibility as long as the cacheDirectory option isn't used.

(It would be possible to implement caching synchronously but I suppose there are performance concerns. We don't need caching when using enhanced-require so I left caching to be async.)

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,7 +22,6 @@ var transpile = function(source, options) {
 };
 
 module.exports = function(source, inputSourceMap) {
-  var callback = this.async();
   var result = {};
   // Handle options
   var defaultOptions = {
@@ -50,6 +49,7 @@ module.exports = function(source, inputSourceMap) {
   this.cacheable();
 
   if (cacheDirectory) {
+    var callback = this.async();
     return cache({
       directory: cacheDirectory,
       identifier: cacheIdentifier,
@@ -63,6 +63,6 @@ module.exports = function(source, inputSourceMap) {
   }
 
   result = transpile(source, options);
-  return callback(null, result.code, result.map);
+  this.callback(null, result.code, result.map);
 
 };


jscheid:
Whoops, totally missed #60! Sorry about that. I guess @camshaft and I have the same problem and approached it from different angles. I don't really care which of the two PRs gets merged, it looks like both will do the trick so feel free to close this.

Dashed:
This PR seems more simpler than the other one.

camshaft:
the other one does support sync caching though. i could go either way as well.

Couto:
The problem with the other one, was that the PR had bad timing since I was rewriting a lot of stuff.
I'm not familiar enough with enhanced-require so I do have to ask... Does caching provide a benefit?

jscheid:
How about merging this one then, seeing that it's pretty straightforward, and leaving the other one for when somebody finds time to review it?

Couto:
@jscheid I think that's a good point, and sorry for taking so long on this matter.
I've merged the PR with the develop branch, can you give a try? If you find no errors with it, I'll make a release with this PR

jscheid:
@Couto thanks, will do -- switch over from our fork to your develop now. I'll let you know how it goes.

jscheid:
@Couto seems to work fine!

Couto:
I've merged this. You can find it on npm with version 5.3.0 
Thanks everyone. If you find any error, please open an issue :)

I'm not familiar with enhaced-required, if caching is something that benefits its process, please tell me, so that I can open an issue to rework the first PR.

PR #85, opened 2015-07-15
Don't override existing items in sourcesContent if any

s-panferov:
The problem appears when babel-loader is used in a chain with another loader, e.g. [awesome-typescript-loader](https://github.com/s-panferov/awesome-typescript-loader) which provides existing source map with own sourcesContent.

In this case we need to leave the items in the array as-is. I've tested it locally, it worked.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,7 +13,7 @@ var transpile = function(source, options) {
   var code = result.code;
   var map = result.map;
 
-  if (map) {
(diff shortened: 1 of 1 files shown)


Couto:
I've just published this on npm with the version 5.3.2! 
Thanks for the PR :)

s-panferov:
@Couto thanks!

PR #94, opened 2015-08-10
Remove `babel-core` from devDependencies and peerDependencies.

### LINKED ISSUE #47: peerDependencies is deprecated
In short, `peerDependencies` should probably be removed:

See for context: https://github.com/npm/npm/issues/6565

SpaceK33z:
Just having it as a dependency is enough; [grunt-babel](https://github.com/babel/grunt-babel) and [gulp-babel](https://github.com/babel/gulp-babel) also do this.

Also fixes #47.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,6 @@
 npm install babel-loader --save-dev
 ```
 
-__Note:__ [npm](https://npmjs.com) will deprecate [auto-installing of peerDependencies](https://github.com/npm/npm/issues/6565) on the next major release, so required peer dependencies like babel-core and webpack will have to be listed explicitly in your `package.json`.
-
 ## Usage
 
 [Documentation: Using loaders](http://webpack.github.io/docs/using-loaders.html)
--- a/package.json
+++ b/package.json
@@ -9,11 +9,9 @@
     "object-assign": "^3.0.0"
   },
   "peerDependencies": {
-    "babel-core": "*",
     "webpack": "*"
   },
   "devDependencies": {
-    "babel-core": "^5.5.8",
     "expect.js": "^0.3.1",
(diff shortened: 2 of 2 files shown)


gaearon:
Hmm.. I actually like that it used to be a peerDep because it puts the burden of choosing the precise version (and update granularity) of `babel-core` onto the user.

gaearon:
But maybe I'm getting it wrong. @sebmck should it be a `dependency`?

sebmck:
AFAIK it's only in `peerDependencies` so npm will match and allow incompatible versions. No reason for it to be in `devDependencies` though if it's `dependencies` though. cc @Couto

Couto:
If i recall correctly, it's only a devDependency due to some uncorrected experiences.

Personally, I think it should be a peer dependency only, since I believe that a plugin/loader should not dictate which babel version the user should use.

On August 11, 2015 1:30:46 AM GMT+01:00, Sebastian McKenzie <notifications@github.com> wrote:
>AFAIK it's only in `peerDependencies` so npm will match and allow
>incompatible versions. No reason for it to be in `devDependencies`
>though if it's `dependencies` though. cc @Couto 
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/babel/babel-loader/pull/94#issuecomment-129656759

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

gaearon:
>Personally, I think it should be a peer dependency only, since I believe that a plugin/loader should not dictate which babel version the user should use.

+1

SpaceK33z:
Okay, peer dependency is fine with me either; I just want one of the two, babel is 20mb...

PR #102, opened 2015-09-10
Update package.json

### LINKED ISSUE #100: babel-core dependencies
`package.json` has specified 3 versions of dependencies for `babel-core`; this seems to be redundant and calls for a clean-up.

gsklee:
Fix #100

Remove redundant dependency declarations for `babel-core`

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "babel module loader for webpack",
   "main": "index.js",
   "dependencies": {
-    "babel-core": "^5.4.0",
+    "babel-core": "^5.8.23",
     "loader-utils": "^0.2.9",
     "object-assign": "^3.0.0"
   },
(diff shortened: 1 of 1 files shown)


joneslee85:
please close

gsklee:
@joneslee85 It's currently in both `peerDependencies` and `devDependencies ` and I don't understand why.

shinnn:
This is why. https://github.com/npm/npm/issues/6565

gsklee:
@shinnn Then why not just listed it as a `dependencies`?

mischkl:
@gsklee 
> Then why not just listed it as a dependencies?

because then it would be forcing a specific version of babel-core on the users of this package. By having it as a peerDependency with the version "*" it can support any version of babel-core the user has installed. But for testing purposes a specific version is specified in devDependencies.

PR #110, opened 2015-09-29
Include NODE_ENV or BABEL_ENV in the cache identifier

### LINKED ISSUE #108: CacheIdentifier should include NODE_ENV
Building with different values for NODE_ENV can result in different output, so it should be included in the cacheIdentifier.

wgcrouch:
Fixes #108


Couto:
This has landed in the develop branch.
It should be on the next release :)

Issue #112, opened 2015-10-04
.babelrc as cache buster didn't quite work


aaronjensen:
I'm not sure what caused it, but somehow this happened: https://github.com/gaearon/babel-plugin-react-transform/issues/32

Specifically, when making several changes to my .babelrc, the output of old version of the babelrc got associated with the cache identifier of the new version. It's possible that webpack dev server was running while I made my changes, or that something else happened while I was switching branches, I do not know. Either way it was pretty confusing.

aaronjensen:
Pretty sure this was due to #113, I'm going to close it out. Feel free to reopen if you think it's worth investigating further.

PR #113, opened 2015-10-05
Include BABEL_ENV || NODE_ENV in cacheIdentifier

aaronjensen:
babel configs can differ based on BABEL_ENV and NODE_ENV, so cache identifier should be different. This may have been the thing that caused #112


Couto:
Thanks for the PR! 
:+1:

aaronjensen:
np, thanks for the merge.

PR #120, opened 2015-10-22
Change loader: babel-loader => loader: babel

yuyokk:
Hello,

There are some places in README that say  _loader: 'babel'_ and some of them  _loader: 'babel-loader'_.
I think it should be changed to be the same everywhere.
Please let me know if this makes sense.

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ module: {
 
   This loader also supports the following loader-specific option:
 
-  * `cacheDirectory`: When set, the given directory will be used to cache the results of the loader. Future webpack builds will attempt to read from the cache to avoid needing to run the potentially expensive Babel recompilation process on each run. The default value (`loader: 'babel-loader?cacheDirectory'`) will cause the loader to use the default OS temporary file directory.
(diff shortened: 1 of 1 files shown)


Dashed:
If the shorthand reference is going to consistent throughout the docs, I think it's good to mention that `babel-loader` is also a legal (explicit) name to reference: http://webpack.github.io/docs/using-loaders.html#referencing-loaders

yuyokk:
@Dashed Oh, I see.

> Within your webpack configuration object, you'll need to add the babel-loader to the list of modules, like so:

```javascript
module: {
loaders: [
  {
    test: /\.jsx?$/,
    loader: 'babel' // 'babel-loader' is also a legal name to reference
  }
]
}
```

What do you think about this variant?

Dashed:
@yuyokk Yep I think that's sufficient. :+1:

maspwr:
LGTM

I also was initially confused by the discrepancy because I wasn't familiar with webpack's resolution logic.

PR #121, opened 2015-10-22
Make babelrc configurable

### LINKED ISSUE #107: .babelrc location not configurable
Is it true that the babelrc location can not be specified in the loader options? (I guess so because of `delete options.babelrc;` in `index.js`). Why is that the case?

MaartenVerbaarschot:
This fixes #107

### DIFF
--- a/index.js
+++ b/index.js
@@ -48,7 +48,6 @@ module.exports = function(source, inputSourceMap) {
 
   delete options.cacheDirectory;
   delete options.cacheIdentifier;
-  delete options.babelrc;
 
   this.cacheable();
 


thealjey:
this would've beed very useful if it wasn't for [#1886](https://github.com/babel/babel/issues/1886#issuecomment-153729262) (see my comment at the bottom)

Couto:
Deleting the option won't affect much since babel is not expecting any `babelrc` option itself: https://babeljs.io/docs/usage/options/

So I have to ask, are you trying to read options from babelrc or just trying to bust the internal cache ?

Couto:
I've read the linked discussion.
As far as I can tell from the docs, there's no babelrc option, still it seems that nothing breaks by allowing the option, so might as well not forbid it.

It's currently in the master and develop branch, give it a try, I should make a release tomorrow

thealjey:
@Couto https://github.com/babel/babel/blob/master/packages/babel-core/src/transformation/file/options/config.js#L131

Couto:
@thealjey lol someone should update the website :dancer:

thealjey:
@Couto
the problem is that it is currently an on/off switch tells babel whether to search for **.babelrc** or not
there's currently no way to specify a specific file to use

For example, `ESLint` has a **--config** flag (**.eslintrc** is just a convenient default), `JSDoc` has a **--configure** flag, `scss-lint` has a **--config** flag for specifying a configuration file to use.

This is basically the reason why to this day I was not able to use **.babelrc** on my project (but I really want to for obvious reasons).
For the API functions it's not that big a deal, because it's easy to pass a config object to them.
As to the CLI it's not as easy, and not possible in some cases.

I would also just love it if the configuration could be written in the YAML format (just a personal preference).

SamanthaAdrichem:
If this still works, could you add it to the documentation in readme.md?

PR #129, opened 2015-11-02
fix cache directory setting in the test

shinnn:
In this case babel-loader should use OS's temporary directory as `cacheDirectory`, but currently uses `cacheDir` instead. https://github.com/babel/babel-loader/blob/a6f3fc00eb53c302187322ca61d1eebcaa075044/test/cache.test.js#L84


shinnn:
Passed. https://travis-ci.org/babel/babel-loader/builds/88753430

PR #146, opened 2015-11-09
Set source file name relative to options.sourceRoot

### LINKED ISSUE #137: Broken source maps
With new babel release I got flattened folder hierarchy :scream:

## Before: 
Before

## After:
After

Example: https://github.com/vlkosinov/babel-lodaer-source-map-issue-example

eur00t:
Fixes #137 by setting `options.sourceFileName`. If I understand correct, it's a basename by default (see Babel code here: https://github.com/chalbert/babel/blob/9d43776fbebf862e56d059dff117e55823f662b5/packages/babel-core/src/transformation/file/index.js#L161)


Couto:
Thanks for the :+1:

Issue #150, opened 2015-11-12
Runtime Option


arb:
It is not clear in the docs how to use the optional babel-runtime configuration. In the documentation where I think that is supposed to be [here](https://github.com/babel/babel-loader#babel-is-injecting-helpers-into-each-file-and-bloating-my-code), it just uses the es2015 configuration value on the query string. Where do you use `babel-runtime`? I tried this

```
      test: /\.js$/,
      include: [
        Path.resolve(process.cwd(), 'client')
      ],
      exclude: 'node_modules',
      loader: 'babel',
      query: {
        presets: ['es2015'],
        optional: ['runtime']
      }
    }
```

But got an error about Unknown option `optional`. What am I missing?

According to https://github.com/babel/babel-loader/issues/123#issuecomment-152673652, you shouldn't need babel-runtime anymore... but then why do I need to install it per the babel-loader documentation?

Couto:
The README.md is outdated.
The `optional` property was [removed in babel v6](https://github.com/babel/babel/blob/master/CHANGELOG.md#600)

dexbol:
I am a newbie and the documents also confused me . Could anybody update the docs?

Couto:
The documents have been updated, [check](https://github.com/babel/babel-loader#babel-is-injecting-helpers-into-each-file-and-bloating-my-code)

PR #156, opened 2015-11-13
update readme to reflect new babel-plugin-transform-runtime plugin

akrawchyk:
This updates the documentation per https://github.com/babel/babel-loader/issues/150 to reflect the new http://babeljs.io/docs/plugins/transform-runtime/ plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,20 +98,20 @@ module: {
   You can instead require the babel runtime as a separate module to avoid the duplication.
 
   The following configuration disables automatic per-file runtime injection in babel, instead
-  requiring `babel-runtime` and making all helper references use it.
+  requiring `babel-plugin-transform-runtime` and making all helper references use it.
 
-  See the [docs](https://babeljs.io/docs/usage/runtime) for more information.
+  See the [docs](http://babeljs.io/docs/plugins/transform-runtime/) for more information.
 
-  **NOTE:** You must run `npm install babel-runtime --save` to include this in your project.
+  **NOTE:** You must run `npm install babel-plugin-transform-runtime --save` to include this in your project.
 
 ```javascript
 loaders: [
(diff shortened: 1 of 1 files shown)


Dashed:
@akrawchyk it should be noted that `npm install babel-runtime --save[-dev]` is still necessary.

akrawchyk:
@Dashed I tried without and it seemed to work... Maybe I'm wrong but the dependency exists on the `babel-plugin-transform-runtime` package.json here https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-runtime/package.json#L12. Shouldn't that take care of it?

Dashed:
@akrawchyk Unsure what your setup is like, but with the `transform-runtime` plugin, something like:

```js
var promise = new Promise;
```

is transformed to:

```js
var _Promise = require('babel-runtime/core-js/promise')['default'];
var promise = new _Promise();
```

And webpack will complain that `babel-runtime` is not found.

AFAIK, this transformation was built into babel 5.x as an option and then refactored out as a plugin for babel 6.x. Looking at the plugin source code, it doesn't seem to inject/require `babel-runtime`.

----

This is my `webpack.config.js` module entry:
```js
// ...
    module: {
        loaders: [
            {
                test: /\.jsx?$/,
                exclude: /node_modules/,
                loader: 'babel-loader',
                query: {
                    cacheDirectory: true,
                    presets: ['es2015', 'react', 'stage-2'],
                    plugins: ['transform-runtime']
                }
            }
        ]
    },
// ...
```

Dashed:
Looking at the commits for `babel-plugin-transform-runtime` package, it doesn't seem that `babel-runtime` was peer dependency.

akrawchyk:
@Dashed added this dependency back in the note, see #164

PR #164, opened 2015-11-23
fix removed dependency note for babel-runtime usage

akrawchyk:
Per @Dashed comment https://github.com/babel/babel-loader/pull/156#issuecomment-158935495, I removed too much from the docs update in #156. 

This PR adds the `babel-runtime` dependency back as an explicit package that is to be installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ module: {
 
   See the [docs](http://babeljs.io/docs/plugins/transform-runtime/) for more information.
 
-  **NOTE:** You must run `npm install babel-plugin-transform-runtime --save` to include this in your project.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,7 +102,7 @@ module: {
 
   See the [docs](http://babeljs.io/docs/plugins/transform-runtime/) for more information.
 
-  **NOTE:** You must run `npm install babel-plugin-transform-runtime --save` to include this in your project.
```
viktorthang: 1

danez:
Thanks this was already fixed by 66913d6d0451a917ef3223607fa8c585fd60a160

PR #181, opened 2015-12-16
Create cache directory if given directory name does not exist

### LINKED ISSUE #180: Use of cacheDirectory requires the directory to already exist (if using a custom cache directory)
For instance:
```
query: {
    cacheDirectory: 'babel_cache',
},
```

For this config to work, the directory babel_cache must already exist. This makes for not the nicest flow when someone is first cloning and setting up a project.

Since git tracks files, not directories, I have to either commit a dummy file in the babel_cache directory or I need to make sure the directory gets created either by instructing the contributor to do so or by having them run a setup script.

Clearly the first option isn't really a big deal, and it's probably what I will do for now. But it would be better if the directory were automatically created when it doesn't exist instead of throwing an error.

fvargas:
Resolves #180

### DIFF
--- a/lib/fs-cache.js
+++ b/lib/fs-cache.js
@@ -126,6 +126,17 @@ var cache = module.exports = function(params, callback) {
   var directory = (typeof params.directory === 'string') ?
         params.directory :
         os.tmpdir();
+
+  // Create the directory if it does not already exist
+  try {
+    fs.mkdirSync(directory);
+  }
+  catch (ex) {
+    if (ex.code !== 'EEXIST') {
+      throw ex;
+    }
+  }
+
   var file = path.join(directory, filename(source, identifier, options));
 
   return read(file, function(err, content) {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "babel-loader",
-  "version": "6.1.0",
+  "version": "6.1.1",
```
jonathanong: you shouldn't set the version in the `package.json` in PRs. let the maintainers do that

lib/fs-cache.js
```diff
@@ -126,6 +126,17 @@ var cache = module.exports = function(params, callback) {
   var directory = (typeof params.directory === 'string') ?
         params.directory :
         os.tmpdir();
+
+  // Create the directory if it does not already exist
+  try {
```
jonathanong: i'd prefer if we just used the `mkdirp` module instead

jonathanong:
:+1: for this features. @Couto @sebmck what can we do to push this forward?

i'm going to make an alternative PR

PR #189, opened 2015-12-26
Remove webpack peer dependency

tlancina:
As far as  I can tell, it doesn't look like babel-loader uses webpack directly, nor does it depend on a specific version of webpack to work, so I'm not sure there's any reason to explicitly list the peer dependency, which causes npm 2 users to have to download and install webpack locally for every project that uses babel-loader.

Even if you'd like to keep the peer dependency, I think the `*` should be made a specific version since things will potentially break if/when webpack 2 moves out of beta (you don't want to say 'any version will do' if the API for webpack 1 and 2 are different, although I have no idea if they are).

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,8 +11,7 @@
     "object-assign": "^4.0.1"
   },
   "peerDependencies": {
-    "babel-core": "^6.0.0",
-    "webpack": "*"
+    "babel-core": "^6.0.0"
   },
   "devDependencies": {
     "babel-core": "^6.0.0",


Couto:
I don't agree with removing webpack as a peer dependency since this loader is quite useless without it.
Some arguments are [these](https://nodejs.org/en/blog/npm/peer-dependencies/)

However I do agree that we should lock the version, in fact it's quite irresponsible from my part to not have locked it before =\

Do you want to update the PR or can I close it and I do the package.json update?

Couto:
I'm going to close this since the lock down has already been done.
Feel free to keep discussing the removal of webpack as a peer dependency if you feel that it really needs to change! :)

tlancina:
Thanks for the response! I can definitely see the argument for keeping it as a peer dependency, since in theory it makes sense.

But from a practical point of view, you won't ever be using a loader without webpack, and forcing users to install a package as large as webpack locally is often unnecessary because many times [it's already a global install](http://webpack.github.io/docs/tutorials/getting-started/#installing-webpack).

Then again, npm 3 specifically addresses issues with peerDependencies, so I think an acceptable solution on our end is to recommend that users upgrade npm.  Again in practice maybe not the best (since npm 3 install times are also way slower), but I understand keeping the peer dependency to ensure compatibility, especially with webpack 2 around the corner :+1:

PR #197, opened 2016-01-08
updates docs re: transform-runtime

gbrassey:
babel 6 does not include the transform-runtime by default so the webpack.ProvidePlugin polyfill should work

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,9 +176,9 @@ loaders: [
 ```
 
 
-### custom polyfills (e.g. Promise library)
+### transform-runtime & custom polyfills (e.g. Promise library)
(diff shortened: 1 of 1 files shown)


Dashed:
I had initially written this part of the docs with babel v5.x in mind. I had assumed not much has changed when I upgraded to babel 6.

@gbrassey Was this step the only thing missing? `npm install babel-plugin-transform-runtime`

gbrassey:
I moved the message to follow the transform-runtime message. Otherwise, the docs look fine to me

danez:
Closed by #331

PR #208, opened 2016-02-03
Update peerDependencies to include Webpack 2

gaearon:
Webpack 2 does not seem to have changes breaking this loader. In this PR, we explicitly enable the latest beta (since we have to specify exact prerelease versions in semver), as well as the stable 2.x when it comes out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   },
   "peerDependencies": {
     "babel-core": "^6.0.0",
-    "webpack": "^1.0.0"
+    "webpack": "^1.0.0 || 2.0.6-beta || ^2.0.0"
   },
   "devDependencies": {
     "expect.js": "^0.3.1",


gaearon:
Note: I’m specified as the “owner” here but I don’t publish packages so it’s up to the actual maintainers to merge this.

Couto:
I'm just trying to fix the travis error, and I'll publish the package right after.
Thanks for the PR :)

Couto:
Travis is taking ages to start with the builds.
This is already in npm with the version v6.2.2

gaearon:
Thank you for quick response!

adamdicarlo:
webpack 2 is now at 2.0.7-beta. Is it possible to specify "2.x-beta" or some such to avoid npm warning? or is the warning desired?

Couto:
The warning is not exactly desired, but as far I can tell, ranges don't work well with pre-releases.

gaearon:
Yeah, you can’t include all future pre-releases.
It might be easier for you to just use `npm@3` that does not force peerDeps.

PR #215, opened 2016-02-17
Allow `babel-loader` to work with the latest webpack beta

blittle:


Couto:
Thanks a lot for the PR! 
However I've just accepted a similar PR. You can check the already published version on npm (6.2.3)

PR #218, opened 2016-02-22
Permit webpack@2.1.0-beta.0

pyrotechnick:


package.json
```diff
@@ -13,7 +13,7 @@
   },
   "peerDependencies": {
     "babel-core": "^6.0.0",
-    "webpack": "1 - 2 || 2.0.6-beta || 2.0.7-beta"
+    "webpack": "1 - 2 || 2.0.6-beta || 2.0.7-beta || 2.1.0-beta.0"
```
Hatlen: if one adds e.g. `^` before `2.1.0-beta.0` it would also match versions like 2.1.0-beta.1 (which has been released now.

So this line would look like this:
 `"webpack": "1 - 2 || 2.0.6-beta || 2.0.7-beta || ^2.1.0-beta.0"`

(this site was really good for testing out npm version strings http://semver.npmjs.com/)

pyrotechnick: The problem is `^2.1.0-beta.0` also matches `2.1.0` which nobody has tested since it doesn't exist yet.

Hatlen: I see, ok then it might be better to lock the version to this strict version as you do.

pyrotechnick: I understand your sentiment and appreciate your input.

At the moment there's no robust way of targeting a range of pre-releases.

pyrotechnick: `1 - 2` already matches `2.1.0` anyway, so yes, this may as well just be `^2.1.0-beta.0` as you suggest.

Open a new PR with your suggestion and I'll nuke this one :)

pyrotechnick: How about `^1 || ^2.1.0-beta.2`?

pyrotechnick:
Superseded by #219

package.json
pyrotechnick: `^1 || ^2.1.0-beta.2` == `1 || ^2.1.0-beta.2`

That's enough SemVer for me today >.>

pyrotechnick:
Superseded by #220

PR #219, opened 2016-02-22
Permit webpack@2 betas

pyrotechnick:
As suggested by @Hatlen


Hatlen:
:+1:

Couto:
I think there's no need for the `.2` after the beta. `1 || ^2.1.0-beta` should be enough

pyrotechnick:
You're right @Couto. I'll clean all of this up in a new PR

PR #224, opened 2016-03-11
Removed babel-preset-react from README.md 

Ghirro:
In line with https://github.com/babel/babel-loader/issues/213

PR #226, opened 2016-03-17
Remove 'react' preset

jrios:
Remove the 'react' preset from the loader query.

It's not needed for the babel loader to work, and it doesn't get installed when the above ``npm install`` command is run. This could bite people that are not using React and just want the babel loader to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ module: {
       exclude: /(node_modules|bower_components)/,
       loader: 'babel', // 'babel-loader' is also a legal name to reference
       query: {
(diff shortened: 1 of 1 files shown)


jrios:
This addresses the README change to account for this PR: https://github.com/babel/babel-loader/pull/224

gaearon:
I see two more mentions of `react` in README, can you please remove them all?

jrios:
@gaearon Good catch. I think I got them all now.

hzoo:
Ah thanks! Sorry for the delay

Issue #166, opened 2015-11-25
`resolveLoader.root` not being taken into account?


julen:
_(I'm not sure if this issue/question belongs here so apologies if that's not the case)_

I'm using Babel 6 and I'm trying to use babel-loader for code which lives in a 3rd party directory, i.e. it's not a parent folder of _webpack.config.js_. This is accessible to webpack via [`resolve.root`](http://webpack.github.io/docs/configuration.html#resolve-root). 
Besides that, I'm setting `resolveLoader: { root: path.join(__dirname, 'node_modules') }` as [described in the webpack documentation](http://webpack.github.io/docs/configuration.html#module-loaders).

This _seems_ to work as the Babel code is run. However, it complains about the `es2015` preset being missing:

```
Module build failed: Error: Couldn't find preset "es2015" relative to directory "myDirectory"
    at OptionManager.mergePresets (~/node_modules/babel-core/lib/transformation/file/options/option-manager.js:327:17)
    at OptionManager.mergeOptions (~/node_modules/babel-core/lib/transformation/file/options/option-manager.js:287:12)
    at OptionManager.init (~/node_modules/babel-core/lib/transformation/file/options/option-manager.js:416:10)
    at File.initOptions (~/node_modules/babel-core/lib/transformation/file/index.js:190:75)
    at new File (~/node_modules/babel-core/lib/transformation/file/index.js:121:22)
    at Pipeline.transform (~/node_modules/babel-core/lib/transformation/pipeline.js:42:16)
    at transpile (~/node_modules/babel-loader/index.js:14:22)
    at ~/node_modules/babel-loader/lib/fs-cache.js:140:16
    at ReadFileContext.callback (~/node_modules/babel-loader/lib/fs-cache.js:27:23)
    at FSReqWrap.readFileAfterOpen [as oncomplete] (fs.js:324:13)
 @ multi editor
```

(here `myDirectory` is the 3rd party directory).

Of course if I `npm install babel-loader babel-core babel-preset-es2015` in the 3rd party directory it works, however, if I understand it correctly, that defeats the purpose of webpack's `resolveLoader.root`.

Am I missing something? Is there something wrong in my understanding? And more importantly, does anyone have any suggestions on how to proceed? Thanks in advance!

fkling:
Would also like to see a proper fix for this. Using `require.resolve` seems to be workaround:

        query: {
          presets: [
            require.resolve('babel-preset-es2015'),
            require.resolve('babel-preset-react'),
[...]

Issue #179, opened 2015-12-15
Unable to use .babelrc when symlinking to a directory above


EvHaus:
My project's folder structure is like this:

```
parent-js/
    root.js
inner/
    js/
        inner.js
    symlinked-js/
        root.js
    node_modules/
        babel-core/
        ...
    .babelrc
    index.html
```

Where `symlinked-js` is a symlink to the `parent-js` directory up a level.

This directory structure is not compatible with babel-loader with babel 6 but was compatible with babel 5. With this structure, when babel 6 tries to load `./../js/root.js` I get the following failure: 

```
WARNING in ../parent-js/root.js
Module parse failed: /path/to/my/app/inner/node_modules/babel-loader/index.js!/path/to/my/app/inner/js/inner.js Line 1: Unexpected token
You may need an appropriate loader to handle this file type.
| import Something from 'something';
```

So basically, the webpack loader for that file is missing.

I can get around this if I stop using `.babelrc` and set a `query` with all my configurations right on the `loader` config in Webpack. And additionally, I need to install all the same modules again in the root folder. So I end up with this:

```
parent-js/
    root.js
inner/
    js/
        inner.js
    symlinked-js/
        root.js
    node_modules/
        babel-core/
        ...
    .babelrc
    index.html
node_modules/
    babel-core/
    ...
```

Note the double `node_modules` it requires me to have in the root.

Is this change done by design in Babel 6? It makes our life more complicated.

gregvenech:
Having the same issue when using symlinks while other loaders (e.g. style, css, sass) all work fine. Similar to [#149](https://github.com/babel/babel-loader/issues/149) as well.

rybon:
For what it's worth, I managed to solve the issue of loading a parent directory by splitting my Webpack config for files in the current directory and files in the parent directory:

```
    module: {
        loaders: [
            {
                test: /\.(js|jsx)$/,
                loader: 'babel', // this makes use of a .babelrc file in the current directory
                include: path.resolve(__dirname, 'src') // only compile project files, no need to compile node_modules
            },
            {
                test: /\.(es)$/, // using a different file extension, we can register a different loader
                loader: require.resolve('babel-loader'), // load babel in a way that resolves in the parent directory
                include: path.resolve(__dirname, '../my_parent_directory_modules'), // the parent directory we want to load
                babelrc: false, // disable the default .babelrc file in the current directory, although it works as well without this flag
[...]

PR #236, opened 2016-04-06
fix(symlinked resources): configure the babelWD

briandipalma:
By specifying the `babelWD` as being the process cwd we allow preset and
plugin resolution to occur relative to the directory where webpack is
running. This should resolve issues where presets and plugins could not
be resolved.

closes: #166, #179


briandipalma:
This PR depends on a babel PR being merged https://github.com/babel/babel/pull/3454

loganfsmyth:
Babel 7.x now has a `cwd` option that defaults to `process.cwd()` and is used to resolve string-paths plugins/presets/paths passed to Babel via `babel-loader`'s options. Given that, this should be resolved.

PR #238, opened 2016-04-18
Add a distributed transpile using process pool.

Strate:
Hey everyone!
This PR implements a distributed compilation usgin awesome [process-pool](https://www.npmjs.com/package/process-pool). It uses all your CPUs for compilation, which gains a performance boost on initial or cacheless run.

Please tell me you opinion about this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,10 +8,8 @@ var read = require('./lib/helpers/read')();
 var resolveRc = require('./lib/resolve-rc.js');
 var pkg = require('./package.json');
 var path = require('path');
-var ProcessPool = require('process-pool');
-var os = require('os');
+var pool = require('process-pool-singleton');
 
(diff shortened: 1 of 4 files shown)


d3viant0ne:
@Strate Thanks for the pull request but i'm thinking we are going to pass on this.

Firstly, the package this is based off of is dated and less than optimally maintained.
More importantly, there are a significant number of runtime environments that this would need to be tested on to include containerized infrastructure where cores are virtual, further adding to the complexity of the said environment.

I like the idea of `AVA` style process pooling, I just don't feel the loader is the right place for that kind of functionality.

@danez - Any thoughts for or agains before I close this?

danez:
Yes I also think async compilation should be handled by webpack and not by the loaders itself.

PR #245, opened 2016-05-11
Update README docs for cacheDirectory's actual behaviour

sohkai:
I'm not sure exactly when the changes were made to fs-cache that outdated the current docs, but given [this line in fs-cache](https://github.com/babel/babel-loader/blob/master/lib/fs-cache.js#L127), it correctly uses the OS's temp dir regardless of if the query param is `?cacheDirectory` or `?cacheDirectory=true` (and, in fact, using the query property syntax results in the same behaviour as `?cacheDirectory=true`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ module: {
 }
   ```
 
-  or by using the query property:
+  or by using the [query property](https://webpack.github.io/docs/using-loaders.html#query-parameters):
 
   ```javascript
 module: {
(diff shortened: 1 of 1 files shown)


danez:
Thanks for the late reply, this works since this commit was merged https://github.com/webpack/loader-utils/commit/1c75aac162e77351442bcc05e186959315abcea3

going to merge this PR by hand.

danez:
Merged in 8fa7946381212ac1636024c0d402c59fba481cb0

sohkai:
Thanks @danez!

PR #246, opened 2016-05-11
not excluding node_modules in specs because the fixtures never require any

antialias:


loganfsmyth:
What's the downside of having this in there? It's good practice to ignore node_modules.

antialias:
Excluding paths with "node_modules" in them was causing specs to fail for me. The reason for this was not obvious to me at first.

I keep all the npm packages that I git clone in a directory called "node_modules", which was causing the "exclude" rule to prevent _any_ files from going through the loader. Apparently `exclude` looks at the absolute path, not just the path relative to cwd.

Since none of the fixtures require any node_modules, it's not necessary to exclude node_modules in the first place, and it's also good practice to no over-configure specs.

If it were easier to specify, I would argue that it's better practice to set
```js
include: new RegExp(
  escape(`${process.cwd()}${path.sep}`) +
  '(?!' +
  escape(`node_modules${path.sep}`) +
  ')'
)
```
As this only allows paths that are in cwd except for `./node_modules`

antialias:
Hi @loganfsmyth. Can you please close or merge this? I'm trying to clean out my pr queue. Thanks!

danez:
I'm gonna close this, and make the assumption, that when running tests you shouldn't be in a `node_modules` folder.

PR #250, opened 2016-05-22
Include react preset in examples

rattrayalex:
http://stackoverflow.com/questions/33460420/babel-loader-jsx-syntaxerror-unexpected-token

Very confusing to include jsx filename with examples that do not work with jsx. 

Feel free to clean this up & make more explicit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 ## Installation
 
 ```bash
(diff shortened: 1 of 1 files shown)


hzoo:
Thanks but we removed the react preset from the docs since it was confusing https://github.com/babel/babel-loader/pull/226 like you mentioend

PR #256, opened 2016-06-01
fix doc of cacheDirecotry

jacksky007:
When I use 
```javascript
loader: 'babel?cacheDirectory'
```
[this line](https://github.com/babel/babel-loader/blob/master/index.js#L32)
```javascript
var loaderOptions = loaderUtils.parseQuery(this.query)
```
returns an object without the key ```cacheDirectory```.
Its causes [this line](https://github.com/babel/babel-loader/blob/master/index.js#L66)
```javascript
var cacheDirectory = options.cacheDirectory
```
returning ```undefined```, so later [cacheDirectory detection](https://github.com/babel/babel-loader/blob/master/index.js#L74) always fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,12 @@ module: {
 
   This loader also supports the following loader-specific option:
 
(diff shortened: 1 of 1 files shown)


jacksky007:
Sorry, I made a mistake here. The doc is right.

PR #260, opened 2016-06-07
fix options.sourceFileName gennerate bug

creeperyang:
One corner usage:

```
sourceRoot: '/Users/me/work/SugarDev',
filename: '/private/tmp/wow/app.js'

// with previous path.relative
sourceFileName: '../../../../private/tmp/wow/app.js'
```

It seems just okay, but when `webpack` use `webpack.optimize.UglifyJsPlugin` with sourceMap, compilation will run into error. Because `webpack.optimize.UglifyJsPlugin` will try to access source map via `'/private/tmp/wow/app.js'`.

More info see https://github.com/webpack/webpack/issues/2616

So we should fix the `relative`.

`webpack` use `source-map` handle source map, and `source-map` use its own `relative`: https://github.com/mozilla/source-map/blob/master/lib/util.js#L188-L226

Here I just replace `path.relative` with `source-map`'s `util.relative`, and It will fix this error.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,91 @@
+'use strict';
+
+var assign = require('object-assign');
+var babel = require('babel-core');
+var loaderUtils = require('loader-utils');
+var cache = require('./lib/fs-cache.js');
+var exists = require('./lib/helpers/exists')();
+var read = require('./lib/helpers/read')();
+var resolveRc = require('./lib/resolve-rc.js');
+var relative = require('./lib/helpers/relative');
+var pkg = require('./package.json');
+var path = require('path');
+
+var transpile = function(source, options) {
+  var result = babel.transform(source, options);
+  var code = result.code;
+  var map = result.map;
+
+  if (map && (!map.sourcesContent || !map.sourcesContent.length)) {
+    map.sourcesContent = [source];
+  }
+
+  return {
+    code: code,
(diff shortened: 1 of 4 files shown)


d3viant0ne:
@creeperyang - Can you get this rebased and up to date so we can get it reviewed and landed?

creeperyang:
@d3viant0ne okay, I will do it.

shankiesan:
Any reason we can't get this merged? This is causing me some pretty serious issues at the moment!

Massive thanks to @creeperyang for sorting this! I would have never though of this...

NickStefan:
@creeperyang I also would love to see this merged. Is this PR still alive?

d3viant0ne:
@NickStefan - I'll give him a another day to respond and if we don't hear anything, i'll get in in directly.

creeperyang:
Yeah, still alive.

danez:
Merged in 501d60d

danez:
@creeperyang I kind of reverted the change and fixed it in a different way. This PR broke some cases on windows. #334 #339 
I reproduced your initial problem with v6.2.7, and still cannot reproduce it with v6.2.9. If you have problems again, please open a ticket with a reproducible case, but let's hope this is fixed forever now 😇

PR #264, opened 2016-06-28
use babel-preset-2016 in readme

nikolas:


loganfsmyth:
This preset only includes features that were added in the `es2016` spec, so you wouldn't replace `es2015` it would be added alongside it.

PR #267, opened 2016-07-14
Update README: add env vars to cacheIdentifier

dferber90:
While enabling the `cacheDirectory` option I was unsure whether it depends on the environment as well, so I had to look into the source. Would be nice if it was mentioned in the README, so I added it.


danez:
Thanks

PR #273, opened 2016-08-04
add babel-core and preset-2015 to dev dependencies

timse:
currently it is impossible to run tests after forking the `babel-loader` without installing `babel-core` and `babel-preset-2015` by hand. Adding them to the dev-dependencies will fix this. (I saw the dependencies as they are installed in the `before_script` according to the `.travis.yml`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,8 @@
     "webpack": "1 || ^2.1.0-beta"
   },
   "devDependencies": {
+    "babel-core": "^6.0.0",
+    "babel-preset-es2015": "^6.0.0",
     "expect.js": "^0.3.1",
     "istanbul": "^0.4.0",
     "jscs": "^2.5.0",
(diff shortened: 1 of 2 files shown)


danez:
Nice thanks, can you then pls also remove them from the `.travis.yml`. That would be nice

package.json
```diff
@@ -16,6 +16,8 @@
     "webpack": "1 || ^2.1.0-beta"
   },
   "devDependencies": {
+    "babel-core": "^6.0.0",
```
d3viant0ne: If we are going to add it as a `devDependency` which is fine as we only support `^6.0.0` it needs to be removed as a `peerDependency`

d3viant0ne: Which is why it is a `peerDependency`. At what point would someone try and use this plugin without already having installed `babel-core` or had it come along with something else `babel` related?

We really shouldn't be packaging `babel-core` in the webpack loader imo.

If it's explicitly added - you're good
If it's a `dependency` of X - you're good
If for some strange reason neither of the above are true, you get a `peerDependency` warning and need to `npm i`

danez: Oh I was reading you want to get rid of the peer dependency. but yeah if the peer dependency stays then we don't have to add it to dependencies

timse:
sure, will do :)

timse:
done and rebased against latest head

danez:
Thanks.

Issue #275, opened 2016-08-06
Cleanup Code Validation and Reporting


d3viant0ne:
- [x] More robust travis build ( should validate against what Webpack supports ) #278
- [x] Discuss moving to eslint with @danez 
- [x] Execute conversion to eslint using `eslint-config-babel` as a base #281
- [x] Branch protection on `master`.
- [x] Coverage generation & PR reporting via codecov.io #278
- [x] Changelog generation, commit format enforcement #279
- [x] Issue & PR Github Templates - `Webpack standard`
- [x] Contributor Guide - `Webpack standard`

> More to follow as I find it ...

danez:
Oh i never noticed it is using jshint. 👍  for the move to eslint, not sure which config tho, `eslint-config-babel` might be a good start.

d3viant0ne:
I usually try to keep code styles inline with parent repos, so in this case `eslint-config-babel` is a perfect place to start  :)

PR #279, opened 2016-08-07
chore(changelog): add changelog generation

d3viant0ne:
- add conventional-changelog with angular presets
- add commitizen adapter
- rewrite contributing Documentation
- add `generate-changelog` script to write the changelog on release

Relates to #275

### DIFF
--- a/scripts/release/generate-changelog.js
+++ b/scripts/release/generate-changelog.js
@@ -1,0 +1,41 @@
+'use strict'
+const fs = require('fs')
+const addStream = require('add-stream')
+const cl = require('conventional-changelog')
+
+const inStream = fs.createReadStream('CHANGELOG.md')
+const isForce = process.argv.indexOf('--force') !== -1
+
+/**
+ * Creates and prepends the changelog from the latest tag to head.
+ * Passing the arg `--force` will rewrite the entire changelog.
+ */
+
+inStream.on('error', function(err) {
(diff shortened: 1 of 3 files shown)


danez:
👍

scripts/release/generate-changelog.js
```diff
@@ -0,0 +1,41 @@
+'use strict'
+const fs = require('fs')
```
abouthiroppy: Do we need ` semicolon`?

d3viant0ne: Need, no.
Should have, yes. I'll update it in a sec.

abouthiroppy:
Sorry for my late reply.
I checked this pr.

LGTM. Thanks😁

scripts/release/generate-changelog.js
```diff
@@ -0,0 +1,39 @@
...
+  releaseCount: isForce ? 0 : 1
+};
+
+const getOutputStream = () => {
+  return fs.createWriteStream('CHANGELOG.md');
+};
+
+const stream = cl(config).on('error', function(err) {
+    console.error(`An error occurred while generating the changelog:  ${err}`);
+  }).pipe(!isForce && addStream(inStream) || getOutputStream());
```
abouthiroppy: [nit] Could you modify indent?

danez:
We don't have that many releases so I think it is not really necessary to do this.

PR #301, opened 2016-09-18
Use standard cache dir as default cacheDirectory

fson:

- [x] Bugfix

Currently, when `cacheDirectory` is set to `true`, `babel-loader` tries to use the operating system's temporary directory as a cache directory.

However, when running npm scripts as the root user, [npm overrides the `TMPDIR` environment variable](https://github.com/npm/npm/issues/4531). This causes the cache files to be created in the project folder itself, for example when using Docker: facebookincubator/create-react-app#483.

**What is the new behavior?**

Use the [common cache directory](https://github.com/avajs/find-cache-dir), `./node_modules/.cache/babel-loader`, as the default cache directory (when the `cacheDirectory` setting is enabled).

- [x] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ module: {
 
   This loader also supports the following loader-specific option:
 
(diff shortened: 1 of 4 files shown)


danez:
Thanks that looks perfect.

I will have a look at babel-register too, as it uses a similar cache strategy and maybe we can put that into a `babel-cache` package.

MoOx:
Any ETA for this? :)

MoOx:
This PR looks great. Anything missing to get this merged?

hzoo:
Thanks for the help, sorry for the long wait

PR #318, opened 2016-10-27
fallback to os.tmpdir() if no cachedir found

### LINKED ISSUE #317: 6.2.6 fails to find cache dir when used from CLI app

**Webpack Version:**
1.13.2

**Babel Core Version**:
6.17.0

**Babel Loader Version**:
6.2.6

Linux

**Current behavior:**
I'm using webpack and the babel-loader from a set of command line utilities wrapping our build process, development server and unit test runner. We are using the cacheDirectory option (set to true) to speed up babel transpilation during development. After upgrading to 6.2.6, this fails at https://github.com/babel/babel-loader/blob/master/lib/fs-cache.js#L131, because findCacheDir returns null.

Digging into the code reveals that findCacheDir looks (via pkg-dir) for a package.json in the current working directory. As the CLI app is globally installed, no package.json can be found.

I'm not sure what the desired cache dir is for globally installed packages. The global node_modules may not be writable due to permissions. Falling back to the old /tmp behavior seems acceptable.

Problem is triggered by invoking webpack with babel-loader from a CLI app, installed globally, with cacheDirectory set to true, running in a directory without any package.json in the ancestry.

* **What is the expected behavior?**
Did not expect build to fail. The code should at least handle the potential null returned by findCacheDir.

Work-arounds are disabling the cacheDirectory (not great for performance during repeated development builds), or manually providing os.tmpdir() to restore the old behavior.

danez:
Fix #317 by falling back to os.tmpdir()

### DIFF
--- a/lib/fs-cache.js
+++ b/lib/fs-cache.js
@@ -127,8 +127,11 @@ var cache = module.exports = function(params, callback) {
   var identifier = params.identifier;
   var directory;
 
-  if (typeof params.directory === 'string') directory = params.directory;
-  else directory = findCacheDir({ name: 'babel-loader' }) || os.tmpdir();
+  if (typeof params.directory === 'string') {
+    directory = params.directory;
+  } else {
+    directory = findCacheDir({ name: 'babel-loader' }) || os.tmpdir();
+  }
 
   var file = path.join(directory, filename(source, identifier, options));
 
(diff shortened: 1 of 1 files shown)


abouthiroppy:
LGTM:+1: 
Thank you :)

PR #254, opened 2016-05-30
pass metadata form babel to a webpack plugin

Ognian:
- add metadataSubscribers option as an array of context functions to call
- pass metadata to metadataSubscribers functions
- add .idea to .gitignore

This is needed when generating messages for translation from modules:
- there is the babel-react-intl plugin, which returns the messages it finds in the module in the metadata[react-int] property. (babel returns not only source code and source map but also the ast and the metadata property )
- since babel loader returns only source code and source map, I extended it to return metadata too. A webpack plugin is now able to aggregate all this metadata...

I will publish such a plugin asap (ReactIntlPlugin)

PR #329, opened 2016-11-11
Feature: Passthru metadata

cogell:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

**What is the new behavior?**

**note**: The following description was lifted from https://github.com/babel/babel-loader/pull/254#issue-157537604 from the original author @Ognian 

- add metadataSubscribers option as an array of context functions to call
- pass metadata to metadataSubscribers functions
- add .idea to .gitignore

This is needed when generating messages for translation from modules:

- there is the babel-react-intl plugin, which returns the messages it finds in the module in the metadata[react-int] property. (babel returns not only source code and source map but also the ast and the metadata property )
- since babel loader returns only source code and source map, I extended it to return metadata too. A webpack plugin is now able to aggregate all this metadata...

- [ ] Yes
- [x] No

* Impact:

**Other information**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,58 @@ require('babel-runtime/core-js/promise').default = require('bluebird');
 require('./app');
 ```
 
+### using `cacheDirectory` fails with ENOENT Error
+
+If using cacheDirectory results in an error similar to the following:
+
+```
+ERROR in ./frontend/src/main.js
+Module build failed: Error: ENOENT, open 'true/350c59cae6b7bce3bb58c8240147581bfdc9cccc.json.gzip'
+ @ multi app
+```
+(notice the `true/` in the filepath)
+
+That means that most likely, you're not setting the options correctly, and you're doing something similar to:
+
+```javascript
+loaders: [
+  {
+    test: /\.jsx?$/,
+    exclude: /(node_modules|bower_components)/,
+    loader: 'babel?cacheDirectory=true'
+  }
+]
+```
+
+That's not the correct way of setting boolean values. You should do instead:
+
+```javascript
+loaders: [
+  {
+    test: /\.jsx?$/,
+    exclude: /(node_modules|bower_components)/,
+    loader: 'babel?cacheDirectory'
+  }
+]
+```
+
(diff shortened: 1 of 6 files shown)


danez:
Thanks for the PR. I cannot see any description on what you are changing and why. Could you elaborate a little bit more?

danez:
Oh I see there is #254

danez:
There are a lot of unneeded commits in this PR, can you clean that up?

cogell:
@danez cleaned up commits and added the description from https://github.com/babel/babel-loader/pull/254 into the top PR comment

Ognian:
@danez @cogell what is now the state of this pr? What can I do to help?

mbrevda:
@Ognian seems like there is a merge conflict. You may want to clean that up.

Ognian:
I'll see what I can do, but it won't be before the end of next week

Ognian:
moving to #396 since I have no write access to vogel repo...

Ognian:
will create a new pr set up on the new version... and link it to this one...

danez:
I have to say I'm not sure if this should be merged into babel-loader, this is very specific to the react-intl usecase but the focus of babel-loader is simply making webpack use babel to parse js files. It might be better if the babel plugin would write its data to the disk and then another loader could simply pickup this data.

Ognian:
@danez sorry but I totally disagree. Maybe at the moment there is only one babel plugin (although I remember at least another one) extracting meta information and passing it thru to webpack, but actually this is a very basic requirement:
If babel returns 3 arguments: code, map and metadata and webpack can be called with those same 3 arguments than babel-loader should also support them.
 Your suggested solution produces not only one file but one file for each source file, producing actually a duplicated source tree for the meta information, and this is just filling up the disk with junk, since webpack compresses this meta information a lot (not to mention that writing to disks slows down the whole thing too). 

At the moment I'm rewriting the tests since you are using now ava. Please tell me if you are considering to merge this otherwise it is useless to write tests for something you will not merge...

Ognian:
ok moved this to #398

PR #334, opened 2016-11-20
Relative paths

josephst:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Right now, relative paths are calculated with POSIX in mind. When run on a Win32 platform, the relative path will fall back to creating an absolute path. In particular, line 23 in `lib/utils/relative.js` causes an error on platforms where paths are delimited by `\` (the `/` is not found, so index is -1, so `aPath` is immediately returned). 

**What is the new behavior?**
When presented with a Win32 style path, `relative.js` will call Node's `path.relative` to calculate a relative path.

- [ ] Yes
- [x] No

* Impact:

**Other information**:

### DIFF
--- a/src/utils/relative.js
+++ b/src/utils/relative.js
@@ -1,3 +1,5 @@
+const path = require("path");
+
 /**
 * Make a path relative to a URL or another path.
 * Borrowed from https://github.com/mozilla/source-map/blob/master/lib/util.js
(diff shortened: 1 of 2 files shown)


josephst:
I need to work on the failing tests

josephst:
Note: tests fail on Node 0.10. path.relative.win32 does not exist so relative.js falls back to current method, which does not work with Windows paths.

abouthiroppy:
@hzoo @danez I think that support of node v1.0.0 is unnecessary. What do you think?

danez:
Closing this in favor of #344
I reverted the fix from #260 and approached it differently, which should not have the same issue on windows, I also added AppVeyor to run the tests on windows.

Issue #327, opened 2016-11-07
Recover from EROFS


frontsideair:
**I'm submitting a bug report**
**I'm submitting a feature request**

**Webpack Version:**
1.13.2

**Babel Core Version**:
6.17.0

**Babel Loader Version**:
6.2.7

**Please tell us about your environment:**
Linux

**Current behavior:**
Fails if target `node_modules` directory is read-only

**Expected/desired behavior:**
It should recover and should disable caching

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem along with a gist/jsbin of your webpack configuration.**
- Make `node_modules` read-only (in my case, it's generated this way by a build tool)
- Run webpack-dev-server (in my case, create-react-app's npm start)

* **What is the expected behavior?**
It should run, without caching

* **What is the motivation / use case for changing the behavior?**
I know this is not the common use case, but out build process depends on Nix package manager, which fetches and builds node dependencies for our project in a read-only directory. When I try to run dev server, it fails since `node_modules/.cache` is read-only. I want it to recover as best as it can.

danez:
Thanks for reporting. 
We currently already have a fallback to the tmp-directory if no `node_modules` directory cannot be find. We could do the same if it is readonly. Would that be satisfy your usecase?

frontsideair:
I think it would work, and it would be better than my suggestion of no caching. Thanks!

frontsideair:
Is there any update or timeline on this?

frontsideair:
I spared some time to fix this myself, can you check it out?

frontsideair:
Awesome, thanks! Now we need a release so create-react-app can use it and they can make a release.

PR #338, opened 2016-11-24
fix cache directory not writable (#327)

frontsideair:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Cannot recover if cache directory is not writable.

**What is the new behavior?**
Switches to tmpdir if cache directory is not writable.

- [ ] Yes
- [x] No

### DIFF
--- a/src/fs-cache.js
+++ b/src/fs-cache.js
@@ -125,35 +125,36 @@ module.exports = function(params, callback) {
   const identifier = params.identifier;
   let directory;
 
-  if (typeof params.directory === "string") {
-    directory = params.directory;
-  } else {
-    directory = findCacheDir({ name: "babel-loader" }) || os.tmpdir();
+  try {
+    if (typeof params.directory === "string") {
+      directory = params.directory;
+      mkdirp.sync(directory);
+    } else {
+      directory = findCacheDir({ name: "babel-loader", create: true });
+    }
+  } catch (e) {
+    // Make sure the directory exists and is writable.
+    directory = os.tmpdir();
   }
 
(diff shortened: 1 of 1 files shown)


src/fs-cache.js
```diff
@@ -125,35 +125,36 @@ module.exports = function(params, callback) {
...
-  if (typeof params.directory === "string") {
-    directory = params.directory;
-  } else {
-    directory = findCacheDir({ name: "babel-loader" }) || os.tmpdir();
+  try {
+    if (typeof params.directory === "string") {
+      directory = params.directory;
+      mkdirp.sync(directory);
+    } else {
+      directory = findCacheDir({ name: "babel-loader", create: true });
```
fson: If `findCacheDir` doesn't find the package root directory, it doesn't throw, only returns `null`. Can you handle that case below? If directory is `null` we should fall back to ` os.tmpdir()`.

frontsideair: Yes, but passing `create: true` makes `findCacheDir` create the directory, which will force it to throw if it doesn't exist.

It would make sense to turn this into a test case.

fson: Only if it can find a parent directory with `package.json`: https://github.com/avajs/find-cache-dir/blob/7df8ba743347945d1f6d51a281fe6776d656b104/index.js#L17-L19

If `pkgDir.sync(dir)` returns `null`, `findCacheDir` will also return `null` regardless of the `create` option.

+1 for adding a test case.

danez:
merged with #346

PR #346, opened 2016-12-11
Fallback for cache directories

### LINKED ISSUE #327: Recover from EROFS

danez:
Follow up of #338
Fixes #327 

@frontsideair I took your commit and modified it. It now correctly falls back if the directory already exists but is not writable. (Before it was only doing the fallback if the directory did not yet exist, as `mkdirp` is only running into an error if it tries to create the directory.) Is this okay, does it work for your usecase?


frontsideair:
Hi, sorry for late feedback. I'm trying this now.

frontsideair:
Seems to be working for me.

PR #365, opened 2017-01-30
document globalOptions

### LINKED ISSUE #364: Documentation for globalOptions
Would you mind if I make a PR adding documentation for `globalOptions` to README?

anilanar:
Fixes #364 

`globalOptions` is useful when a babel plugin needs non-serializable options (such as functions, callbacks), thus it's worth mentioning in the README.

- docs update

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# babel-loader 
-[![NPM Status](https://img.shields.io/npm/v/babel-loader.svg?style=flat)](https://www.npmjs.com/package/babel-loader) 
-[![Build Status](https://travis-ci.org/babel/babel-loader.svg?branch=master)](https://travis-ci.org/babel/babel-loader) 
+# babel-loader
+[![NPM Status](https://img.shields.io/npm/v/babel-loader.svg?style=flat)](https://www.npmjs.com/package/babel-loader)
+[![Build Status](https://travis-ci.org/babel/babel-loader.svg?branch=master)](https://travis-ci.org/babel/babel-loader)
 [![Build Status](https://ci.appveyor.com/api/projects/status/vgtpr2i5bykgyuqo/branch/master?svg=true)](https://ci.appveyor.com/project/danez/babel-loader/branch/master)
 [![codecov](https://codecov.io/gh/babel/babel-loader/branch/master/graph/badge.svg)](https://codecov.io/gh/babel/babel-loader)
   > Babel is a compiler for writing next generation JavaScript.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
-# babel-loader 
-[![NPM Status](https://img.shields.io/npm/v/babel-loader.svg?style=flat)](https://www.npmjs.com/package/babel-loader) 
-[![Build Status](https://travis-ci.org/babel/babel-loader.svg?branch=master)](https://travis-ci.org/babel/babel-loader)
```
anilanar: My editor automatically deletes trailing spaces, but I guess that's desirable for everyone so I'm keeping them.

danez:
global options is not supported as of webpack 2, so I would right now not add it to the documentation.
You can always use the query syntax:
```js
module: {
  loaders: [
    {
      test: /\.js$/,
      loader: 'babel-loader',
      query: {
        presets: ['es2015']
      }
    }
  ]
}
```

anilanar:
@danez As I've mentioned:

> globalOptions is useful when a babel plugin needs non-serializable options (such as functions, callbacks)

For example, I have some babel plugins to collect (and not transform) occurrences of a particular expression type in all of my javascript files and process collected expressions. One of those plugins collect `translate(x)` expressions where `translate` has a binding to a particular module. Webpack's own parser is not so great at doing that, and webpack parser kicks in after all transpiling is complete. So it's desirable to do it using a babel plugin.

Tricky part is how to send collected expressions to the outside world. For that, I pass a function callback as an **option** to the babel plugin. The **only** way of achieving that using `babel-loader` is by using global options.

For Webpack2, I'm hoping that `rules.options` is not stringified using `JSON.stringify` before it's passed to the loader. If it is (in which case functions cannot be used as an option), then webpack2 introduces `LoaderOptionsPlugin` to handle global options. If there are existing webpack2 documentation for `babel-loader` I can update them accordingly.

danez:
For webpack2 the `rules.options`  is directly forwarded https://github.com/webpack/loader-runner/blob/master/lib/LoaderRunner.js#L346

The Readme for babel-loader is for both webpack 1 & 2 right now, but we want to release a new major version soon for webpack 2 only.

danez:
Could you add a note that this form of defining options is deprecated and in webpack2 on should use `rules.options` (https://webpack.js.org/configuration/module/#rule-options-rule-query)?

danez:
Thank you, I added the note about webpack 2.

PR #368, opened 2017-02-02
Add env query parameter.

### LINKED ISSUE #283: Setup BABEL_ENV for Webpack2 via loader
Need the ability to pass the desired environment name to babel via `babel-loader`.

I have such `.babelrc`:

```
{
  "presets": [
    "es2015",
    "stage-0",
    "react"
  ],
  "env": {
    "webpack2": {
      "presets": [
        ["es2015", {"modules": false}],
        "stage-0",
        "react"
      ],
    }
  }
}
```

I'm starting webpack via run script which written with es6 modules. But if I start script with `BABEL_ENV=webpack2` then it requires rewriting all scripts to commonjs style.  

Furthermore I should rewrite all app, cause use Relay with babelRelayPlugin.

So the better way to use Webpack2 with passing env value via `babel-loader` settings.

moimael:
Allow to override BABEL_ENV/NODE_ENV at loader-level. Useful for isomorphic applications which have separate babel config for client and server.

fix #283

- [ ] Tests for the changes have been added (not sure if needed here, but i'd like to hear what maintainers think)

- [x] Feature

See #283 

**What is the new behavior?**
If `forceEnv` is set as loader option, use that instead of BABEL_ENV/NODE_ENV

- [x] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# babel-loader 
-[![NPM Status](https://img.shields.io/npm/v/babel-loader.svg?style=flat)](https://www.npmjs.com/package/babel-loader) 
-[![Build Status](https://travis-ci.org/babel/babel-loader.svg?branch=master)](https://travis-ci.org/babel/babel-loader) 
+# babel-loader
+[![NPM Status](https://img.shields.io/npm/v/babel-loader.svg?style=flat)](https://www.npmjs.com/package/babel-loader)
+[![Build Status](https://travis-ci.org/babel/babel-loader.svg?branch=master)](https://travis-ci.org/babel/babel-loader)
 [![Build Status](https://ci.appveyor.com/api/projects/status/vgtpr2i5bykgyuqo/branch/master?svg=true)](https://ci.appveyor.com/project/danez/babel-loader/branch/master)
 [![codecov](https://codecov.io/gh/babel/babel-loader/branch/master/graph/badge.svg)](https://codecov.io/gh/babel/babel-loader)
   > Babel is a compiler for writing next generation JavaScript.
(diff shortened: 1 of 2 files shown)


moimael:
Can someone have a look ?

abouthiroppy:
@moimael Sorry for the late reply. It looks good to me. cc: @babel/webpack

danez:
Thanks you

tleunen:
@moimael Could you explain how to use it?
I get the changes you've made in the PR, then try to add the option, but when my app loads, I'm getting this error now:
```
Module build failed: ReferenceError: [BABEL] /src/index.js: Unknown option: base.forceEnv.
```

Here's the loader:
```
{
    test: /\.jsx?$/,
    exclude: /node_modules/,
    loader: 'babel-loader',
    options: {
        forceEnv: 'local'
    }
},

**Comment by moimael** [comment]
How did you install it ? Using file:// ? Did you rebuild it before ?

**Comment by tleunen** [comment]
Manually added the missing piece, but obviously I missed `delete userOptions.forceEnv;`.

It's good now, thx :)

**Comment by chrisvasz** [comment]
This feature isn't working for me. Looking at the code, it looks like this change only affects the `cacheIdentifier`, but doesn't actually change the env for the Babel transform operation. I had to add something like this to `index.js` to make babel-loader actually *use* the specified environment:
```
var env = userOptions.forceEnv || process.env.BABEL_ENV || process.env.NODE_ENV;
// ...
var tmp = process.env.BABEL_ENV;        // added
process.env.BABEL_ENV = env;            // added
result = transpile(source, options);
process.env.BABEL_ENV = tmp;            // added
```

I'm happy to send this in a PR, but I wanted to make sure I'm not missing something first.

moimael:
@chrisvasz It is working well for me, so I'm a bit surprised. How do you use it ? I might very well have missed something since it's my first contribution to anything babel related.

tleunen:
Same for me. it works well.

I don't think it makes sense to update BABEL_ENV. The custom value is passed to babel when compiling, which is why it works.

chrisvasz:
@tleunen I'm not seeing where the custom value gets passed into Babel. `userOptions.forceEnv` is used as part of the cacheIdentifier, then deleted before being assigned into `options`. I dug in because it wasn't working for me... I'm happy to be told that I'm doing something wrong, but looking at the code, I'm not sure how this *could* work.

tleunen:
Oh, you're right, I missed the stringify function when reading the code.

dKab:
How it's intended to be set? Like this?

    let path = require('path');

    module.exports = {
        entry: './src/index.js',
        output: {
            path: path.resolve(__dirname, '../dist'),
            filename: 'bundle.js',
            publicPath: '/'
        },
        module: {
            rules: [
                {
                    test: /\.(js|jsx)$/,
                    exclude: /node_modules/,
                    use: {
                        loader: 'babel-loader',
                        options: {
                            forceEnv: {
                                BABEL_ENV: 'client'
                           }
                        }
                     }
                }
            ]
        }
    };

moimael:
More like `forceEnv: 'client'`

PR #371, opened 2017-02-07
Update loader-utils dependecy to ^0.2.16

leonaves:
Anything below 0.2.16 for loader-utils causes issues with webpack 2 (https://github.com/webpack/webpack/issues/3060#issuecomment-249382018 and I have replicated the issue)

- [x] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Webpack build fails with the error:

```
Module build failed: Error: parseQuery should get a string as first argument
    at Object.parseQuery (/path/to/node_modules/loader-utils/index.js:68:9)
    at Object.module.exports (/path/to/node_modules/babel-loader/lib/index.js:79:35)
```

**What is the new behavior?**

Webpack build succeeds

- [ ] Yes
- [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "main": "lib/index.js",
   "dependencies": {
     "find-cache-dir": "^0.1.1",
-    "loader-utils": "^0.2.11",
+    "loader-utils": "^0.2.16",
     "mkdirp": "^0.5.1",
     "object-assign": "^4.0.1"
   },


danez:
Thank you, although this new version will be picked up anyway as we use carret. But it also makes sense to increase the version.

deitch:
Ah, spent the last little bit tracking this down. When does this get released to npm?

danez:
At some point, for now just remove your node_modules and install again.

leonaves:
I'm pretty sure if you have another dependency that hard requires 0.2.15 or below, npm will happily install the lower version and be content that it has satisfied the dependency of babel-loader.

deitch:
> I'm pretty sure if you have another dependency that hard requires 0.2.15 or below, npm will happily install the lower version and be content that it has satisfied the dependency of babel-loader

Yes. For now, I basically removed babel-loader, added loader-utils@0.2.16 explicitly, then added babel-loader. It probably will blow up again if I do another install, but hopefully last long enough until this is published.

PR #379, opened 2017-02-14
Apply forceEnv option to Babel transformation

chrisvasz:
As per the discussion in #368, this PR applies the `forceEnv` option to the Babel transformation.

Bug fix

As discussed [here](https://github.com/babel/babel-loader/pull/368#issuecomment-279742678), the `forceEnv` option applies to the cacheIdentifier, but does not actually apply to the Babel transformation.

**What is the new behavior?**
The `forceEnv` option now applies to the Babel transformation as well.

- [x] No

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -78,6 +78,7 @@ module.exports = function(source, inputSourceMap) {
   const globalOptions = this.options.babel || {};
   const loaderOptions = loaderUtils.parseQuery(this.query);
   const userOptions = assign({}, globalOptions, loaderOptions);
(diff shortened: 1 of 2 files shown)


tleunen:
What about with the `cacheDirectory` flag?

chrisvasz:
What about it? I tried to think of how `cacheDirectory` would factor into this change, but I couldn't think of a way immediately. `cacheIdentifier` definitely makes sense.

tleunen:
`process.env.BABEL_ENV` is still not set when `cacheDirectory` is `true`.
The transpilation is done inside the cache. See `transform: transpile`.

PR #391, opened 2017-02-22
Bump loader-utils and use getOptions

existentialism:
Fixes warning:

```sh
DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.
```

Ref: https://github.com/webpack/loader-utils/issues/56

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "main": "lib/index.js",
   "dependencies": {
     "find-cache-dir": "^0.1.1",
-    "loader-utils": "^0.2.16",
+    "loader-utils": "^1.0.2",
     "mkdirp": "^0.5.1",
     "object-assign": "^4.0.1"
   },
--- a/src/index.js
+++ b/src/index.js
@@ -76,7 +76,7 @@ module.exports = function(source, inputSourceMap) {
 
   // Handle options
   const globalOptions = this.options.babel || {};
-  const loaderOptions = loaderUtils.parseQuery(this.query);
+  const loaderOptions = loaderUtils.getOptions(this);
   const userOptions = assign({}, globalOptions, loaderOptions);
   const defaultOptions = {
     inputSourceMap: inputSourceMap,
--- a/yarn.lock
+++ b/yarn.lock
@@ -2627,6 +2627,14 @@ loader-utils@^0.2.11, loader-utils@^0.2.16:
     json5 "^0.5.0"
     object-assign "^4.0.1"
 
+loader-utils@^1.0.1:
+  version "1.0.2"
+  resolved "https://registry.yarnpkg.com/loader-utils/-/loader-utils-1.0.2.tgz#a9f923c865a974623391a8602d031137fad74830"
+  dependencies:
+    big.js "^3.1.3"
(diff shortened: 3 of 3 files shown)


danez:
This breaks as the new loader-utils version drops support for node < 4

Timer:
Considering that Node 4 is the oldest supported version of Node (0.10 and 0.12 being dead), one could argue it's not a breaking change.

kaumac:
+1 
this will hopefully fix #392

gaearon:
Can we release babel-loader@7 without waiting for Babel 7? Doesn't seem like they have to track each other.

Nilegfx:
+1

d3viant0ne:
@danez - Nothing related to Webpack 2.x supports node < 4. Agree with @Timer, I wouldn't consider this breaking as support for .10 / .12 was never there to begin with. This has been handled as a `patch` version for all loaders requiring the same change across the `webpack-contrib` org.

hassox:
I've been trying to get a new project up and this issue is so far very difficult to solve. This plugin is now at odds with most other loaders and loaderUtils does not support parseQuery anymore :( I'd love to see this (or something similar) get merged in soon :D

mrtnbroder:
Please release this asap, it's causing issues on our app.

blade254353074:
...We are waiting for this release

danez:
The deprecation warning is now fixed in v7.0.0-alpha.2

mrtnbroder:
Really? You've put this into an alpha-version instead of simply bumping to the next version?

danez:
The updated loader-utils version dropped support for node <4 which babel-loader 6 still supports and we only drop node versions in major versions. So there was no other way we wanted to do this.
It is also only a deprecation notice, nothing to worry about.

We will release 7.0.0 asap, as @gaearon correctly said the loader is not coupled to the babel 7 process.

mgol:
@danez `loader-utils` is still tested in Node 4: https://github.com/webpack/loader-utils/blob/master/.travis.yml so it seems they should support it?

danez:
but not for node <4

mgol:
Ah, sorry, I read it incorrectly.

corysimmons:
Been through a couple walls of text. Using Node 7.4.0 with latest webpack and babel-loader. Still getting this error. Can someone summarize what's wrong and what needs to be done?

Timer:
@corysimmons if you're using `latest` of `babel-loader`, that's your problem. You need to be on `next`.

If you are on next, go ahead and run node with the `--trace-deprecation` flag to figure out exactly where the deprecation is coming from.

corysimmons:
`yarn add -D babel-loader@next`?

Timer:
Yup, if you want to be more explicit @7.0.0-beta.1

corysimmons:
Thanks, works. 👍

Vanuan:
Does babel-loader@next work with babel-core 6?

danez:
@Vanuan Yes babel-core does work with 6.x

PR #409, opened 2017-03-06
Add prettier

danez:
Enables prettier

//cc @vjeux Not many drastic changes here, but in case you want to have a look

//cc @hzoo @yavorsky How should we handle eslint? I disabled one rule `arrow-parens` as it was interfering, should we extend `eslint-config-prettier` in our config?

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,10 @@
     "eslint": "^3.8.1",
     "eslint-config-babel": "^6.0.0",
     "eslint-plugin-flowtype": "^2.25.0",
+    "husky": "^0.13.2",
+    "lint-staged": "^3.3.1",
     "nyc": "^10.0.0",
+    "prettier": "^1.2.2",
     "react": "^15.1.0",
     "react-intl": "^2.1.2",
     "react-intl-webpack-plugin": "^0.0.3",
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -88,5 +92,15 @@
       "src/**/*.js"
     ],
     "babel": "inherit"
+  },
+  "lint-staged": {
+    "src/**/*.js": [
+      "prettier --trailing-comma all --write ",
+      "git add"
+    ],
+    "test/**/*.test.js": [
```
yavorsky: Should we also include `test/helpers/createTestDirectory.js` ?

yavorsky:
eslint-config-prettier is a good idea. I'm going to add it for https://github.com/babel/babel-preset-env/pull/183.

But seems like adding
```json
"arrow-parens": "off",
"indent": "off"
```
is enough to make babel-config prettier-friendly.

package.json
hzoo: Do we need to do all of these or can we just run it on all .js files?

src/fs-cache.js
```diff
@@ -57,13 +56,12 @@ const write = function(filename, result, callback) {
   const content = JSON.stringify(result);
 
   return zlib.gzip(content, function(err, data) {
-    if (err) { return callback(err); }
+    if (err) return callback(err);
```
hzoo: I like the braces but ok

package.json
yavorsky: @hzoo I think the only reason why can't do this is that running on all .js files will affect fixtures

src/fs-cache.js
vjeux: Note that this is not prettier doing. Prettier will keep the {} as they were inputted. But, if there are braces, the expression will be on its own line, if there are no braces, the expression is inline like it is here.

existentialism:
It might be fine to just remove the few style rules we have in eslint-config-babel?

src/fs-cache.js
hzoo: Ok I didn't think so

PR #420, opened 2017-03-14
Fixed reset of BABEL_ENV when options.forceEnv is used

nikopavlica:

- [x] Bugfix

After using 'forceEnv' option, BABEL_ENV is set to "undefined" (string), causing
unintended behaviour of babel running in different env.

**What is the new behavior?**
BABEL_ENV is correctly reset to previous value OR deleted if it wasn't set before

- [ ] Yes
- [x] No

**Other**
Some new serial tests added to test the environment variable clearing. Serial is needed
because otherwise, parallel execution interfere with setting/clearing of the variable.
Maybe these tests could be done with less config - feel free to change/update them :)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -46,7 +46,7 @@ const transpile = function(source, options) {
   try {
     result = babel.transform(source, options);
   } catch (error) {
-    if (forceEnv) process.env.BABEL_ENV = tmpEnv;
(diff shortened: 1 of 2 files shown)


danez:
Good catch. thanks for fixing.

nikopavlica:
Took a while to figure out why such "random" behaviour of my app :-)

No probs and thanks for merge.

PR #427, opened 2017-03-23
Make sure .babelrc is a file, not a directory

hirokiosame:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If there is a _directory_ called `.babelrc` in the repo root or it's parent, the build breaks with an error: `Module build failed: Error: EISDIR: illegal operation on a directory, read` stemming from `find (/Users/hiroki/Documents/Gits/babel-loader/lib/resolve-rc.js:21:12)`

**What is the new behavior?**
If the detected `.babelrc` is not a file, it ignores it and moves on.

- [ ] Yes
- [x] No

* Impact:

**Other information**:
There is logic in `babel-core` in `build-config-chain.js` that checks if `.babelrc` exists but neglects to check if it's a file. I will make a PR fixing it there as well.

### DIFF
--- a/src/utils/exists.js
+++ b/src/utils/exists.js
@@ -14,7 +14,7 @@ module.exports = function(cache) {
 
     if (!filename) { return false; }
 
-    cache[filename] = cache[filename] || fs.existsSync(filename);
+    cache[filename] = cache[filename] || (fs.existsSync(filename) && fs.statSync(filename).isFile());
 
     return cache[filename];
   };


loganfsmyth:
We can certainly throw a better error, but I don't think silently ignoring it makes sense?

hirokiosame:
If there's a directory that is coincidentally named `.babelrc` in it's parent, that has nothing to do with the build (or perhaps not even Babel), does it make sense to throw an error that will prevent the build from happening?

That is the case I am currently encountering. When executing Babel on a shared server I do not have admin control in, it fails because someone created a `.babelrc` folder.

I think the particular part of the code I updated is not only concerned with whether `.babelrc` exists, but more so that it's _readable_.

abouthiroppy:
https://github.com/babel/babel/pull/5535

loganfsmyth:
Hmm, that's a fair point, if a little gross. I guess I can see the utility of this.

PR #433, opened 2017-04-04
Support global options in LoaderOptionsPlugin

egoist:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

babel-loader 7 does not accept global options defined in `webpack.LoaderOptionsPlugin`

**What is the new behavior?**

Just like babel-loader 6

- [ ] Yes
- [x] No

**Other information**:

I don't know if you want to support this in babel-loader 7 though, feel free to close this if you don't.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -101,7 +101,8 @@ module.exports = function(source, inputSourceMap) {
   const filename = webpackRemainingChain[webpackRemainingChain.length - 1];
 
   // Handle options
-  const loaderOptions = loaderUtils.getOptions(this) || {};
+  const query = loaderUtils.getOptions(this) || {};
+  const loaderOptions = Object.assign({}, this.options.babel, query);
   const defaultOptions = {
     metadataSubscribers: [],
     inputSourceMap: inputSourceMap,


abouthiroppy:
CC @danez

egoist:
Hmm, can we merge this? 😅

Couto:
I was under the impression that the LoaderOptionsPlugin exists only to help people migrate to webpack 2.

What use case are you trying to support by passing the options to the plugin instead of the loader? :)

egoist:
Oh, I just found a solution to my use case, so it's fine even without this being merged now 😊 

But I guess this should be supported, it's useful for sharing options too.

danez:
I had the same impression as @Couto. 

@egoist Can you share how you fixed it and what the usecase is?

egoist:
@danez I need to configure babel for both babel-loader and vue-loader, so now I need to do it twice, which seems to be the recommended way in webpack 2. So it's fine to close this since babel-loader@7 only wants to support webpack 2.

Couto:
@egoist I'm not familiar with vue-loader but I'm guessing that you can always declare a variable with the babel configuration and share it between the two, since webpack config files are simple JavaScript files :) (Just an idea though)

PR #456, opened 2017-05-30
fix(resolve-rc): look for babel in package.json

mastilver:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

**What is the new behavior?**

- [ ] Yes
- [x] No

* Impact:

**Other information**:

### DIFF
--- a/src/resolve-rc.js
+++ b/src/resolve-rc.js
@@ -12,25 +12,34 @@ const read = require("./utils/read")({});
 
 const cache = {};
 
-const find = function find(start, rel) {
-  const file = path.join(start, rel);
+const find = function find(start) {
+  const babelRcFile = path.join(start, ".babelrc");
 
-  if (exists(file)) {
-    return read(file);
+  if (exists(babelRcFile)) {
+    return read(babelRcFile);
+  }
+
+  const packageJsonFile = path.join(start, "package.json");
+
+  if (exists(packageJsonFile)) {
+    const parsedPackageJson = require(packageJsonFile);
+
+    if (parsedPackageJson.babel) {
+      return JSON.stringify(parsedPackageJson.babel);
+    }
   }
 
   const up = path.dirname(start);
   if (up !== start) {
     // Reached root
-    return find(up, rel);
+    return find(up);
   }
 };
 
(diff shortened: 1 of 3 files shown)


danez:
Awesome thank you very much. 
By looking at the code I noticed that `babelrc.js` is not supported (which is coming with babel 7).
Instead of further modifying this code, what do you think about using https://github.com/tleunen/find-babel-config? It already supports the new file. And would make the babel-loader code more easy by outsourcing this part.

evilebottnawi:
@danez 
> We would still need to do the caching.
Now it is not work as expected. Step:
1. Run `webpack` in watch mode.
2. Just edit config file - rebuilding process not started.

How to solve:
1. Use `this.addDependency` for configuration file (`.babelrc` and etc), how it is do `postcss-loader` https://github.com/postcss/postcss-loader/blob/master/lib/index.js#L93

mastilver:
@evilebottnawi I think what @danez was saying is for performance sake `find-babel-config` should have caching

What you mention is a different issue and from my point of view is not an issue
`webpack` only watch "code" files not config files, if you update your `.babelrc` you are required to restart `webpack`

evilebottnawi:
@mastilver @danez This is fundamentally wrong and breaks the logic of watching and caching.
Each loader using config files should be add their to watch graph.

From webpack channel as say @sokra:
```shell
sokra [4:54 PM] 
configuration files -> loaders should add these files to their dependencies (with `this.addDependency`). dependencies' mtimes are stored by the cache-loader
```

Also it is break loader before, example `cache-loader`, actually loaders using before `babel-loader` a lot more and many of their also can improve own cache. All this makes the cache invalid and is a wrong look at watch graph.

mastilver:
@evilebottnawi I take back what I said but can you open an issue about it, this PR is not related to your issue

danez:
Thanks this was merged with d8b73c04fc0b984a3c80d9b918f45ddc95fe9893

mastilver:
@danez Thank you for taking care of it :) , don't have much time atm

ismay:
@danez This hasn't yet been released on npm right?

PR #460, opened 2017-06-08
Support watching of config files

evilebottnawi:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Loader don't watch configuration files. Also don't adding configuration breaks other loader before `babel-loader`. And provide invalidate config after change.

**What is the new behavior?**

Watch on config files.

- [ ] Yes
- [x] No

* Impact:

**Other information**:
Ref: https://github.com/babel/babel-loader/pull/456#issuecomment-307143909
Ref: https://github.com/babel/babel-loader/pull/456#issuecomment-307161324

### DIFF
--- a/src/utils/exists.js
+++ b/src/utils/exists.js
@@ -1,22 +1,21 @@
-const fs = require("fs");
 /**
  * Check if file exists and cache the result
  * return the result in cache
  *
  * @example
- * var exists = require('./helpers/fsExists')({});
- * exists('.babelrc'); // false
+ * var exists = require('./helpers/fsExists');
+ * exists(require('fs'), '.babelrc'); // false
  */
-module.exports = function(cache) {
-  cache = cache || {};
+module.exports = function(fileSystem, filename) {
+  if (!filename) return false;
 
-  return function(filename) {
-    if (!filename) return false;
+  let exists = false;
 
-    cache[filename] =
-      cache[filename] ||
-      (fs.existsSync(filename) && fs.statSync(filename).isFile());
+  try {
+    exists = fileSystem.statSync(filename).isFile();
+  } catch (ignoreError) {
+    return false;
+  }
 
-    return cache[filename];
-  };
+  return exists;
 };
--- a/src/utils/read.js
+++ b/src/utils/read.js
@@ -1,22 +1,16 @@
-const fs = require("fs");
 /**
  * Read the file and cache the result
  * return the result in cache
  *
  * @example
- * var read = require('./helpers/fsExists')({});
- * read('.babelrc'); // file contents...
+ * var read = require('./helpers/fsExists');
+ * read(require('fs'), '.babelrc'); // file contents...
  */
-module.exports = function(cache) {
-  cache = cache || {};
+module.exports = function(fileSystem, filename) {
+  if (!filename) {
+    throw new Error("filename must be a string");
+  }
 
-  return function(filename) {
-    if (!filename) {
-      throw new Error("filename must be a string");
-    }
-
-    cache[filename] = cache[filename] || fs.readFileSync(filename, "utf8");
-
-    return cache[filename];
-  };
+  // Webpack `fs` return Buffer
+  return fileSystem.readFileSync(filename).toString("utf8");
 };
--- a/src/index.js
+++ b/src/index.js
@@ -2,11 +2,12 @@ const babel = require("babel-core");
 const loaderUtils = require("loader-utils");
 const path = require("path");
 const cache = require("./fs-cache.js");
-const exists = require("./utils/exists")();
+const exists = require("./utils/exists");
 const relative = require("./utils/relative");
-const read = require("./utils/read")();
+const read = require("./utils/read");
 const resolveRc = require("./resolve-rc.js");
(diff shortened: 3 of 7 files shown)


src/utils/read.js
```diff
@@ -7,16 +7,10 @@ const fs = require("fs");
  * var read = require('./helpers/fsExists')({});
  * read('.babelrc'); // file contents...
  */
-module.exports = function(cache) {
-  cache = cache || {};
```
danez: Removing this cache makes `babel-loader` got to the file-system for every file that it compiles, which seems not to be a good idea.
Or can you explain the reason for this change?

evilebottnawi: @danez now cache provide old config file when changes happens, i known what it is not best, I'll try to find a solution, also related to `resolve-rc.js`. While it's possible that caching such a place is just the wrong solution.

evilebottnawi: @danez main problems not in `restarting watch when the config changes`, main problems in watch graph (absence in dependencies configuration), it is break loader caches before `babel-loader`

evilebottnawi:
@danez Sad but I could not find a solution. I don't know what can we do with this, but I feel this is a strong problem. It seems that caching a configuration is a bad idea.

sokra:
You can use the filesystem provided via `this.fs` to the loader. It cached but invalidated by watching.

danez:
@sokra That is awesome. I was just searching on how to find out what change is triggering the loader run, so that we could invalidate the the cache, but that is even easier.

evilebottnawi:
@danez friendly ping

src/utils/exists.js
```diff
@@ -1,22 +1,21 @@
-const fs = require("fs");
 /**
  * Check if file exists and cache the result
  * return the result in cache
  *
  * @example
- * var exists = require('./helpers/fsExists')({});
- * exists('.babelrc'); // false
+ * var exists = require(require('fs'), './helpers/fsExists')({});
```
danez: This exmaple is wrong I think ;)

src/utils/exists.js
```diff
@@ -1,22 +1,21 @@
...
 
-  return function(filename) {
-    if (!filename) return false;
+  let exists = false;
 
-    cache[filename] =
-      cache[filename] ||
-      (fs.existsSync(filename) && fs.statSync(filename).isFile());
+  try {
+    exists = fileSystem.statSync(filename).isFile();
```
danez: Can we do `fs.existsSync(filename)` here instead of exception catching. Try catch will make v8 deoptimize this function.

src/utils/read.js
```diff
@@ -1,22 +1,16 @@
-const fs = require("fs");
 /**
  * Read the file and cache the result
  * return the result in cache
  *
  * @example
- * var read = require('./helpers/fsExists')({});
- * read('.babelrc'); // file contents...
+ * var read = require(require('fs'), './helpers/fsExists')({});
```
danez: Also funny require

src/utils/read.js
```diff
@@ -1,22 +1,16 @@
...
-    if (!filename) {
-      throw new Error("filename must be a string");
-    }
-
-    cache[filename] = cache[filename] || fs.readFileSync(filename, "utf8");
-
-    return cache[filename];
-  };
+  // Webpack `fs` return Buffer
+  return fileSystem.readFileSync(filename, "utf8").toString();
```
danez: As the encoding is specified, we do not need to do `.toString()` here.

src/utils/exists.js
evilebottnawi: @danez webpack fs doesn't `fs.existsSync`, because `exists` is deprecated

src/utils/read.js
evilebottnawi: @danez `fileSystem.readFileSync` in webpack fs doesn't support `second` argument and always return `Buffer`, i try to investigate this

src/utils/exists.js
danez: I see, maybe webpack would be open to add `existsSync` as it was undeprecated in 7.0.

evilebottnawi: @danez hm, your are right, in theory we can create issue and add comment in code, and when issue was resolved change code?

src/utils/read.js
evilebottnawi: @danez seems some fs function don't support more than one argument https://github.com/webpack/enhanced-resolve/blob/master/lib/CachedInputFileSystem.js#L197.

src/utils/exists.js
evilebottnawi: @danez btw, `webpack` have many places where use `statSync` and this does not cause any problems with performance, i think this can be left

src/utils/read.js
evilebottnawi: Practically
`fileSystem.readFileSync(filename).toString("utf8");` === `fileSystem.readFileSync(filename, "utf8");`.
https://github.com/nodejs/node/blob/master/lib/fs.js#L602

evilebottnawi:
@danez friendly ping

danez:
Thank you very much this looks good.

Issue #93, opened 2015-07-27
Relative paths in source maps


byelims:
I'm trying to switch from jsx-loader. My `webpack.config.js` is like this:

```js
module.exports = {
  entry: './src/javascript/main.jsx',
  output: {
    devtoolModuleFilenameTemplate: '[resource-path]',
    path: './public/js/',
    filename: 'bundle.js'
  },
  module: {
    loaders: [
      { test: /\.jsx$/, loaders: ['jsx'] }
    ]
  }
};
```

This will produce relative paths in source maps.

```js
{
  sources: ["./src/javascript/main.jsx"]
}
```

If I use babel-loader with the same config, it will produce absolute paths.

```js
{
  sources: ["/full/path/src/javascript/main.jsx"]
}
```

Are there any methods to solve this problem?

jonjaques:
Babel supports JSX natively, is there any reason for not using that functionality? I heard React's JSX transform is being deprecated.

waeljammal:
I second this, it's not nice having full paths..

nielssj:
Agreed, I experience the same issue. 
The incorrect paths seem to break the WebStorm debugger.

elektronik2k5:
+1

winseros:
Same here, babel-loader v.6.2.0

mischkl:
I came up with a simple fix that should solve the problem - check it out.

danez:
Is this still a problem?

when I try to reproduce the sources look like this: `"webpack:///./fixtures/constant.js"`

dizel3d:
![image](https://cloud.githubusercontent.com/assets/593571/20785959/290f7fbc-b7b4-11e6-9ad2-7dafd7803552.png)

chrisnicola:
Just noticed this may be a duplicate of the issue I just filed #480

hiradyazdan:
@byelims you probably need to check whether in your .babelrc your babel presets for production env is different (e.g. if using react it's not `react-hmre`) than your development.

PR #481, opened 2017-06-24
Fix for sourcemap file locations

### LINKED ISSUE #480: Breaks absolute file paths for source map filename templates

*Delete what isn't applicable below*

**Webpack Version:**
Tested with 2.x and 3.0.0

**Babel Core Version**:
6.25.0

**Babel Loader Version**:
7.1.0

OSX 10.x

**Current behavior:**

If I look at the values passed to `output.devtoolModuleFilename` for files parsed through `babel-loader` have a non-relative and non-absolute value. Basically the values for `resource`, `resourcePath` and `absoluteResourcePath` are all the same.

Specifically if I have a file at `./src/main.js` I will get

resource: `src/main.js`
resourcePath: `src/main.js`
absoluteResourcePath: `src/main.js`

Files that are not passed through babel-loader have the following information:

resource: `src/main.js`
resourcePath: `./src/main.js`
absoluteResourcePath: `/Usrs/someone/dev/project/src/main.js`

Using `babel-loader` you can debug this by setting `output.devtoolModuleFilename` to a console logging function and viewing the results.

chrisnicola:
babel-loader was setting `options.sourceFileName` to a relative path, but this does not seem to work correctly anymore with either webpack or babel itself and causes incorrect source map file data in webpack.

Setting it to an absolute path seems to fix the problem.

Fixes #480 and possibly also #93.


package.json
```diff
@@ -3,9 +3,9 @@
   "version": "7.1.0",
   "description": "babel module loader for webpack",
   "files": [
-    "lib"
+    "src"
   ],
-  "main": "lib/index.js",
```
chrisnicola: Ah I see, because the build script outputs to `/lib`. I mainly made this change so I could install from my git branch for testing and didn't really think about it. 

That may also explain why it isn't working with Node <8.0. I still have to do some testing of this today, but I'll revert that.

chrisnicola: It looks like using a proper build tarball fixed my issues and everything is working again. Thanks for that.

chrisnicola:
@michael-ciniawsky I'm unclear what you're referring to. Do you mean this shouldn't be assigned to `process.cwd()`? https://github.com/chrisnicola/babel-loader/blob/38ed9a3db991a3f098d48f1e4ac734dc6d74fc15/src/index.js#L124

michael-ciniawsky:
According to @sokra `sourceRoot` should be `''` and webpack itself takes care of it. Note that I ask for `postcss-loader` (CSS), so it might vary here, definitely needs triage, but `process.cwd()` seem to be odd :)

chrisnicola:
@michael-ciniawsky you are correct that sourceRoot doesn't need to be set. However for the absoluteResourcePath to not simple return the filename the sourceFileName must be set to `options.filename`

At least that's what my testing showed.

@danez is there anything else I need to do for this PR?

chrisnicola:
@danez are any other changes required for this.

@michael-ciniawsky I've tested this and it does work with sourceRoot set to '' or unset, but it still requires that the sourceFileName be set to the absolute path of the file. It doesn't seem to matter what the sourceRoot is set to.

chrisnicola:
@danez hey I'm sure your busy, so I'd be happy to take care of any work necessary to get this merged in. Is there anything I can do?

chrisnicola:
Still hoping to see this merged sometime. Right now we are packaging the fix manually to make sure source maps work.

Issue #477, opened 2017-06-23
Warning if transpiling ES modules to commonjs


jakearchibald:
Some presets, such as 'es2015' will transpile ES6 modules to commonjs. This works fine, but it prevents webpack performing treeshaking, meaning bundles could be unnecessarily larger.

Is there any reason users would want to transpile to commonjs at this point in the pipeline? If not, babel-loader could attempt to change the default here, or at least display a warning if it detects this kind of transpilation occurring.

elrumordelaluz:
@jakearchibald based on the [babel-preset-es2015 documentation](https://github.com/babel/babel/tree/master/packages/babel-preset-es2015#options) you could set something like:
```
options: {  
  presets: [ 
    [ 'es2015', { modules: false } ] 
  ]         
}
```
> Setting this to false will not transform modules

Hope it works,

jakearchibald:
Yeah, I used the same workaround, but since babel-loader is built for webpack, and webpack handles modules better than commonjs (eg tree shaking), it should be the default.

Unless there's a reason users would want to transpile to commonjs before passing to webpack?

Couto:
Imho, this would break the predictability of both tools. 

Webpack and Babel are already hard to set up, and messing with the tools' defaults will probably cause a lot of unaware mistakes, since Babel recommends to turn the option off when you want to delegate the modules to another tool, instead of having the tools to mess with the options behind the scenes.

I do believe, however, that we could update the README.md examples to include that option, since like you stated, it does make sense to delegate the modules resolution to Webpack.

jakearchibald:
@Couto 

> this would break the predictability of both tools … messing with the tools' defaults will probably cause a lot of unaware mistakes

Right, but you have to balance that with the unaware mistake of losing tree shaking without warning.

Sure, you could add something to the readme along the lines of "btw a lot of the common defaults do the wrong thing", but a runtime warning sounds a lot more useful.

Couto:
If we don't mess with the option behind the scenes, and it just throws a warning, I think that's a cool feature.

Would you be willing to do a PR?

jakearchibald:
I'm really just learning webpack right now & creating issues as I find them as a user. I don't know how plugins are made (nor I'm I familiar with the internals of Babel), so it might be a while before I can ramp up to a decent PR. Eg, I'm sure there's a standard way to do logging, but I have no idea.

But if no one knowledgeable is willing to take it, I guess I could get to it eventually.

[...]

PR #485, opened 2017-06-27
feat: Added warning when modules is not false (#477)

wardpeet:

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

#477 

**What is the new behavior?**
When webpack is version 2 or higher and we're using es2015 or env preset
We throw a warning saying that webpack can't treeshake correctly

- [ ] Yes
- [x] No

* Impact: Users with not having the right configs on webpack2 will get a warning
* Github Issue(s) this is regarding: #477 

**Other information**:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,6 +9,10 @@ const resolveRc = require("./resolve-rc.js");
 const pkg = require("../package.json");
 const fs = require("fs");
 
+const presetsToCheckForModules = ["es2015", "env"];
+const modulesWarningMsg = `We've noticted the option modules isn't set to false,
+which disables treeshaking.`;
+
 /**
  * Error thrown by Babel formatted to conform to Webpack reporting.
  */
@@ -147,6 +151,28 @@ module.exports = function(source, inputSourceMap) {
 
   if (options.sourceFileName === undefined) {
     options.sourceFileName = relative(options.sourceRoot, options.filename);
+  }
+
+  if (this.version > 1 && options.presets) {
+    options.presets.forEach(preset => {
+      let name = preset;
+      let opts = {};
+
+      // if we have extra options put it in opts
+      if (Array.isArray(preset)) {
+        name = preset[0];
+        opts = preset[1];
+      }
+
+      if (
+        presetsToCheckForModules.indexOf(name) > -1 &&
+        (!opts ||
+          (opts && !opts.hasOwnProperty("modules")) ||
+          (opts && opts.modules))
+      ) {
+        this.emitWarning(new Error(modulesWarningMsg));
+      }
+    });
   }
 
   const cacheDirectory = options.cacheDirectory;
--- a/src/index.js
+++ b/src/index.js
@@ -9,9 +9,8 @@ const resolveRc = require("./resolve-rc.js");
 const pkg = require("../package.json");
 const fs = require("fs");
 
-const presetsToCheckForModules = ["es2015", "env"];
-const modulesWarningMsg = `We've noticted the option modules isn't set to false,
-which disables treeshaking.`;
+// we keep track of each preset config
+const emitCache = new Map();
 
 /**
  * Error thrown by Babel formatted to conform to Webpack reporting.
@@ -165,12 +164,20 @@ module.exports = function(source, inputSourceMap) {
       }
 
       if (
-        presetsToCheckForModules.indexOf(name) > -1 &&
-        (!opts ||
-          (opts && !opts.hasOwnProperty("modules")) ||
-          (opts && opts.modules))
+        ["es2015", "env", "latest"].indexOf(name) > -1 &&
+        (!emitCache.has(options.presets) &&
+          (!opts ||
+            (opts && !opts.hasOwnProperty("modules")) ||
+            (opts && opts.modules)))
       ) {
-        this.emitWarning(new Error(modulesWarningMsg));
+        emitCache.set(options.presets, true);
+        this.emitWarning(
+          new Error(
+            `\n\n⚠️  Babel Loader\n
+It looks like your Babel configuration specifies a module transformer. Please disable it.
+See https://babeljs.io/docs/plugins/preset-${name}/#optionsmodules for more information.`,
+          ),
+        );
       }
     });
   }
--- a/src/index.js
+++ b/src/index.js
@@ -174,7 +174,8 @@ module.exports = function(source, inputSourceMap) {
         this.emitWarning(
           new Error(
             `\n\n⚠️  Babel Loader\n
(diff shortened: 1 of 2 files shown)


wardpeet:
it's not done yet. I have to add a check so it only fires the warning once and add some more tests to make sure it works 100%.

Note that I only check modules on es2015 and env preset.

src/index.js
```diff
@@ -9,6 +9,10 @@ const resolveRc = require("./resolve-rc.js");
 const pkg = require("../package.json");
 const fs = require("fs");
 
+const presetsToCheckForModules = ["es2015", "env"];
+const modulesWarningMsg = `We've noticted the option modules isn't set to false,
```
existentialism: Nit: noticted  -> noticed

src/index.js
```diff
@@ -149,6 +153,28 @@ module.exports = function(source, inputSourceMap) {
...
+        opts = preset[1];
+      }
+
+      if (
+        presetsToCheckForModules.indexOf(name) > -1 &&
+        (!opts ||
+          (opts && !opts.hasOwnProperty("modules")) ||
+          (opts && opts.modules))
+      ) {
+        this.emitWarning(new Error(modulesWarningMsg));
```
michael-ciniawsky:
```
this.emitWarning(`
⚠️ Babel Loader\n
We've noticed that module transpilation (`option.modules`) is enabled.
webpack supports ES2015 Modules natively and module transpilation disables tree shaking.
It is recommended to disable it (https://github.com/babel/path/to/explanation)
`)
```

src/index.js
```diff
@@ -149,6 +153,28 @@ module.exports = function(source, inputSourceMap) {
...
+      let opts = {};
+
+      // if we have extra options put it in opts
+      if (Array.isArray(preset)) {
+        name = preset[0];
+        opts = preset[1];
+      }
+
+      if (
+        presetsToCheckForModules.indexOf(name) > -1 &&
```
michael-ciniawsky: Just add the {Array} here :)

src/index.js
michael-ciniawsky: The message can of course be improved in terms of spelling, but the `message` convention should be like above :)

michael-ciniawsky: loaderwarning

src/index.js
```diff
@@ -9,6 +9,10 @@ const resolveRc = require("./resolve-rc.js");
 const pkg = require("../package.json");
 const fs = require("fs");
 
+const presetsToCheckForModules = ["es2015", "env"];
```
michael-ciniawsky: Are these two really to only ones ?

danez: we also could add `latest` although I think it is deprecated.

src/index.js
wardpeet: thanks not sure this is the message we want to show though 😛

src/index.js
wardpeet: I don't know how many there are to be honest..

src/index.js
wardpeet: great thanks!

nathanhammond:
Notes from the meeting:

---

Your goal is to allow the syntax to parse, but enable Webpack to do the transformation.

We don't want to tie this to the preset string, instead we want to ensure that Babel knows that the babel syntax plugin is enabled and the transform plugin is not being run.

That doesn't mean that we won't accept this in present state to alleviate the worst problems, just that we have a target for the future.

wardpeet:
@nathanhammond perhaps there is a way to see if babel has ran the module transformation?

src/index.js
```diff
@@ -149,6 +152,36 @@ module.exports = function(source, inputSourceMap) {
...
+        (!emitCache.has(options.presets) &&
+          (!opts ||
+            (opts && !opts.hasOwnProperty("modules")) ||
+            (opts && opts.modules)))
+      ) {
+        emitCache.set(options.presets, true);
+        this.emitWarning(
+          new Error(
+            `\n\n⚠️  Babel Loader\n
+It looks like your Babel configuration specifies a module transformer. Please disable it.
```
danez: Can we maybe give a little bit more information? Like "... . Please disable it, in order to allow webpack to create optimized bundles." Or "... . This disables tree shaking in webpack and will produce larger bundles. Please disable it."

wardpeet: updated

danez: Did you maybe forget to push? :)

wardpeet: yes sorry :)

src/index.js
```diff
@@ -149,6 +152,37 @@ module.exports = function(source, inputSourceMap) {
...
+      }
+
+      if (
+        ["es2015", "env", "latest"].indexOf(name) > -1 &&
+        (!emitCache.has(options.presets) &&
+          (!opts ||
+            (opts && !opts.hasOwnProperty("modules")) ||
+            (opts && opts.modules)))
+      ) {
+        emitCache.set(options.presets, true);
```
danez: should this be `name` instead of `options.presets`? Otherwise can you explain why you are setting the array as key?

wardpeet: The reason why I chose options.presets over name is that I cache a whole collection instead of just 1 preset.

Let me explain a bit more clearly. Lets say you have different load rules which use a different set of rules.
```
{
  test: /\.js$/,
  use: {
    loader: 'babel-loader',
    options: {
      presets: ['es2015'],
      plugins: [require('babel-plugin-transform-object-rest-spread')]
    }
  }
},

{
  test: require.resolve('legacy-module'),
  use: {
    loader: 'babel-loader',
    options: {
      presets: ['es2015'],
    }
  }
}
```
Now I get 2 warnings because I've setup 2 load rules instead of just 1 as the presets are different

src/index.js
```diff
@@ -149,6 +152,37 @@ module.exports = function(source, inputSourceMap) {
...
+      if (Array.isArray(preset)) {
+        name = preset[0];
+        opts = preset[1];
+      }
+
+      if (
+        ["es2015", "env", "latest"].indexOf(name) > -1 &&
+        (!emitCache.has(options.presets) &&
+          (!opts ||
+            (opts && !opts.hasOwnProperty("modules")) ||
```
loganfsmyth: I'd avoid `opts.hasOwnProperty` as an instance property. Also you never validate that `opts` is an object, so this would throw if someone passed an invalid config instead of passing through to `babel-core`'s validation. I'd go with

```
const PRESET_WHITELIST = new Set(["es2015", "env", "latest"]);

// ...

const { modules } = typeof opts === "object" && opts || {};

if (PRESET_WHITELIST.has(name) && modules !== false) {
  this.emitWarning // ...
}
```

wardpeet:
closing in favor of #529

PR #494, opened 2017-07-24
babelrc-related bugfixes

Mitranim:
* don't try to load missing babelrc (`false` check was too narrow)
* don't swallow non-FS errors when checking if file exists

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -112,7 +112,7 @@ module.exports = function(source, inputSourceMap) {
   const loaderOptions = loaderUtils.getOptions(this) || {};
   const fileSystem = this.fs ? this.fs : fs;
   let babelrcPath = null;
-  if (loaderOptions.babelrc !== false) {
+  if (loaderOptions.babelrc) {
     babelrcPath = exists(fileSystem, loaderOptions.babelrc)
       ? loaderOptions.babelrc
       : resolveRc(fileSystem, path.dirname(filename));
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -112,7 +112,7 @@ module.exports = function(source, inputSourceMap) {
   const loaderOptions = loaderUtils.getOptions(this) || {};
   const fileSystem = this.fs ? this.fs : fs;
   let babelrcPath = null;
-  if (loaderOptions.babelrc !== false) {
```
danez: That completely changes the behaviour for when not setting babelrc option.

danez:
Can you describe what problem you actually try to resolve?

Mitranim:
The problematic scenario is very basic, I'm surprised other people don't complain.

* `webpack@3.3.0` + `babel-loader@7.1.1`
* `.babelrc` in project root
* rudimentary webpack config with `use: {loader: 'babel-loader'}` and no plugin options
* run `webpack --watch`, touch a file → webpack blows up and dies

Specifically: `babel-loader` attempts to check for existence of `babelrc` with the `undefined` name (not a string, just `undefined`), passing it to Webpack's virtual filesystem, which assumes that the file name is a string and produces a synchronous exception.

Checking for `false` is naive. By default, most users won't specify a `babelrc` location, meaning the option is `undefined`. The change accommodates the most common use case while keeping the same general logic.

zythum:
because enhanced-resolve updated api. if exist
path is undefined will threw error  https://github.com/webpack/enhanced-resolve/pull/100

danez:
Merged in #495

Issue #503, opened 2017-08-20
Error: fileSystem.statSync is not a function in 7.1.2


jamro:
**I'm submitting a bug report**

**Webpack Version:**
1.15.0

**Babel Core Version**:
6.26.0

**Babel Loader Version**:
7.1.2

**Please tell us about your environment:**
OSX 10.12.6

**Current behavior:**
I'm getting following error when building with gulp and webpack:
```
[12:52:08] Using gulpfile gulpfile.js
[12:52:08] Starting 'build'...
[12:52:09] Version: webpack 1.15.0

ERROR in ./src/entry.js
Module build failed: TypeError: fileSystem.statSync is not a function
    at module.exports (node_modules/babel-loader/lib/utils/exists.js:7:25)
    at find (node_modules/babel-loader/lib/resolve-rc.js:13:9)
    at Object.module.exports (node_modules/babel-loader/lib/index.js:113:132)
[12:52:09] Finished 'build' after 1.17 s

Gulp ran 1 task in 1.17 s
```
Here is my Gulp script:

```
module.exports = function (gulp, config, plugins) {
  return function () {
    return gulp.src(config.sources.concat(config.lib))
      .pipe(plugins.webpack({
        entry: config.entry,
        output: {
          filename: 'lib.js',
          library: 'lib',
          libraryTarget: 'var'
        },
        module: {
          loaders: [{
            test: /.js$/,
            loader: 'babel-loader',
            exclude: /node_modules/,
            query: {
              presets: ['es2015']
            }
          }]
        },
        node: {
          fs: 'empty'
        }
      }))
      .pipe(gulp.dest(config.dist + 'js/'));
  };
};
```

In version 7.1.1 of babel-loader the problem did not occur. The reason is probably because all errors in `exists.js` are swallowed in 7.1.1:

```
  try {
    exists = fileSystem.statSync(filename).isFile();
  } catch (e) {}
```

When I remove try & catch in 7.1.1, I'm getting the same error as in 7.1.2.

ibiBgOR:
Same on Windows 7 Professional SP1

antennaio:
Confirmed. I'm getting the exact same error.

dnbr2002:
I am having the same issue.  

Webpack Version:
 1.14.1

Babel Core Version:
 ^6.26.0

Babel Loader Version:
 ^7.1.1

Please tell us about your environment:
 Win7x64 SP1


[...]

PR #504, opened 2017-08-21
fix https://github.com/babel/babel-loader/issues/503

sergey-chechaev:
BugFix https://github.com/babel/babel-loader/issues/503


danez:
Thanks for the PR. This is wrong and breaks the code. The issue is simply that babel-loader 7 does not work with webpack 1 anymore.

PR #522, opened 2017-10-30
Breaking: Change peerDep to @babel/core

hzoo:
Ok removed inherits since it's not necessary

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,11 @@
 {
   "presets": [
-    ["env", { "loose": true, "targets": { "node": "4.2" } }]
+    ["@babel/preset-env", {
+      "loose": true,
+      "targets": {
+        "node": "4.2"
+      }
+    }]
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -80,9 +79,6 @@
       "text",
       "json"
     ],
-    "require": [
-      "babel-register"
```
hzoo: don't need register either for nyc

.babelrc
```diff
@@ -1,6 +1,11 @@
 {
   "presets": [
-    ["env", { "loose": true, "targets": { "node": "4.2" } }]
+    ["@babel/preset-env", {
```
loganfsmyth: You can leave off the `preset-` if you want.

Also looks like there are a bunch of things in the README that will need updating.

hzoo: ok updated

PR #529, opened 2017-11-04
Add `modules: false` by default for es2015/env/latest presets

### LINKED ISSUE #477: Warning if transpiling ES modules to commonjs

### LINKED ISSUE #521: Set `modules: false` by default on babel-preset-env

**Webpack Version:**
>2.x

**Babel Loader Version**:
7.1.2

**Current behavior:**

I need to manually disable the `modules` feature of Babel like this:

```js
test: /\.js$/,
use: {
    loader: 'babel-loader',
    options: {
        presets: ['env', { modules: false }]
    }
}
```

To let webpack use its own ESM system.

For webpack 2.x and higher, babel-loader can detect automatically that we don't want to use `modules`.

Sean said I had to mention `loaderContext.version = 2` as a way to detect the new webpack version.

Many people don't know this, and **especially** people who upgraded from webpack 1.x to >2.x often don't have this enabled (seen this a lot personally).

One problem with this is that it is a breaking change (or at least potentially), but at the long term I think it's worth it. One less configuration option for most people.

**Open question**: is there a valid use-case where you want to have `modules` on `true` in webpack 2?

cc'ing @hzoo and @TheLarkInn because I know you talked about this a few minutes ago with eachother 🥇 

Related: #519

SpaceK33z:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

See #521.

**What is the new behavior?**

The Babel presets `env`, `latest` and `es2015` will be mutated to add `modules: false` as an option. This also works when you use e.g. `@babel/preset-env` or `@babel/env`.

- [x] Yes
- [ ] No

For people who use babel-loader without setting the `modules` option, there might be small differences in the way webpack handles ES modules.

For people who still want to keep the old behavior, they can explicitly add `modules: true` or e.g. `modules: "amd"`.

**Other information**:

I am not happy at all with the implementation, but it works. Let me know if there is a better way I missed.

I used parts of the code in #485, so credits to @wardpeet!

Fixes #521, fixes #477, fixes #485

### DIFF
--- a/src/cache.js
+++ b/src/cache.js
@@ -80,6 +80,14 @@ const filename = function(source, identifier, options) {
 const handleCache = async function(directory, params) {
   const { source, options = {}, cacheIdentifier, cacheDirectory } = params;
 
+  const file = path.join(directory, filename(source, cacheIdentifier, options));
+
+  try {
+    // No errors mean that the file was previously cached
+    // we just need to return it
+    return await read(file);
+  } catch (err) {}
+
   const fallback =
     typeof cacheDirectory !== "string" && directory !== os.tmpdir();
 
@@ -93,14 +101,6 @@ const handleCache = async function(directory, params) {
 
     throw err;
   }
-
-  const file = path.join(directory, filename(source, cacheIdentifier, options));
-
-  try {
-    // No errors mean that the file was previously cached
-    // we just need to return it
-    return await read(file);
-  } catch (err) {}
 
   // Otherwise just transform the file
   // return it to the user asap and write it in cache


Andarist:
Override shouldnt be done with true but with a module type to match the mentioned presets options

wardpeet:
@SpaceK33z should I abandon my PR?

SpaceK33z:
@wardpeet, yeah this would make your PR unnecessary. Sorry about that, but I did end up using much of your code!

danez:
I would be still in favor of showing a warning similar to `rollup-plugin-graphql` instead of modifying the users configuration. 
It always is a little bit frustrating to me when tools do this, because I guess that people who currently rely on babel producing commonjs will have a hard time understanding why suddenly their _correct_ config does not work anymore and does not do what they expect.

src/index.js
```diff
@@ -152,6 +152,34 @@ module.exports = function(source, inputSourceMap) {
     options.sourceFileName = relative(options.sourceRoot, options.filename);
   }
 
+  if (this.version > 1 && options.presets) {
+    const presetsWithModulesOption = ["es2015", "env", "latest"];
```
michael-ciniawsky:
```js
const presets = [ "es2015", "env", "latest" ]

// not sure if really needed
const name = (name) => name
  .replace(/^@babel\//, "")
  .replace(/^preset-/, "")

options.presets.forEach((preset) => presets.indexOf(name(preset[0])) > -1 
   ? preset[1] !== undefined 
     ? preset[1].modules = false
     : preset[1] = { modules: false }
   : preset
)

michael-ciniawsky: But I agree with @danez [here](https://github.com/babel/babel-loader/pull/529#issuecomment-342196655) that a warning with link(s) to the appropriate docs would be the saner solution. `babel` is configurable and modular by design and a dev simply needs to know how to configure it properly

SpaceK33z:
I messed up this PR by using force push, so I made a new PR #650.

PR #538, opened 2017-11-13
fix compatibility with @babel/core option check when inputSourceMap null

Delagen:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Assertion in @babel/core package when file come from another loader without sourceMap
In my case `val-loader` with only `code` option.
https://github.com/babel/babel/blob/master/packages/babel-core/src/config/option-assertions.js#L71

**What is the new behavior?**
No errors on compilation.

- [ ] Yes
- [x] No

**Other information**:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -127,7 +127,7 @@ module.exports = function(source, inputSourceMap) {
 
   const defaultOptions = {
     metadataSubscribers: [],
-    inputSourceMap: inputSourceMap,
+    inputSourceMap: inputSourceMap || undefined,
     sourceRoot: process.cwd(),
     filename: filename,
     cacheIdentifier: JSON.stringify({


danez:
Sorry for the late reply. So this fails if inputSourceMap is `null`? Do I understand this correctly?

Can you please add a comment describing why we need to default to null, because I already see that in a year someone might remove it because it is not common to default to undefined. :)

Delagen:
@danez Yes. Because babel-core only check for undefined, and throw exception if it null

Delagen:
@danez Yes. Because babel-core only check for undefined, and throw exception if it null

Delagen:
@danez Yes. Because babel-core only check for undefined, and throw exception if it null

Delagen:
@danez Seems latest builds of @babel/core does not affected, so I close it for now.

Delagen:
Again get error Module build failed: Error: .inputSourceMap must be a boolean, object, or undefined

danez:
Thanks

Delagen:
@danez https://github.com/babel/babel/blob/493996e02a3e22ae4e89443f4df106ea6bc3774c/packages/babel-core/src/config/validation/option-assertions.js So it must be omited from params or undefined

danez:
Yeah i noticed that, not sure why though. @loganfsmyth 

This fix fix should work for now.

PR #550, opened 2017-12-07
Allow any version of webpack as a peerDependency

lencioni:
Webpack 4 alpha has been released. In preparation for v4,
it seems reasonable to change this to allow `*`. This should cut
down on maintenance and avoid unnecessarily blocking folks
from updating going forward.

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: peerDependency change

Cannot be used with Webpack 4 alpha

**What is the new behavior?**
Can be used with Webpack 4 alpha

- [ ] Yes
- [x] No

* Impact:

**Other information**:
It would be nice to put out a release of babel-loader that allows any version of webpack and also works with Babel 6.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   },
   "peerDependencies": {
     "@babel/core": "7 || ^7.0.0-beta || ^7.0.0-rc",
-    "webpack": "*"
+    "webpack": ">=2"
   },
   "devDependencies": {
     "@babel/cli": "7.0.0-beta.5",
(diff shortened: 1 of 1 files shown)


danez:
Can you please change this to >= 2 that is what it should be right now, as 1.x is not supported and will not work.

danez:
Thank you very much.

loganfsmyth:
Is using a range like this a good idea? Using `*` is bad practice and I feel like `>=`2 isn't better, future Webpack versions could still easily break things if they wanted. How about `>=2 < 5`?

Andarist:
I second what @loganfsmyth said, it seems pretty much as `*` in disguise. Even specifying <5 is a little bit stretch as we do not have certainty that it will work for webpack@4 - it is not released after all.

PR #574, opened 2018-02-05
Make compressing transforms cache configurable

### LINKED ISSUE #571: What's the idea behind compressing cached output
**Babel Loader Version**:
current (8.0.0-beta.0) but I believe it stays since beginning (or at least 6.x)

**Current behavior:**
When `cacheDirectory` option is set, `babel-loader` stores the cached transforms as compressed files (using zlib).

I'm actually curious what's the idea behind compressing cache? 

The only reason I can see is to make `node_modules` take less place (as `node_modules/.cache/babel-loader` is the default cache dir)

The drawback, however, is spending quite some time compressing and decompressing (not to mention stringifying and parsing prior to that).

Would you be interested in accepting a PR that _at least_ would make compression customisable?

thymikee:
- [x] Tests for the changes have been added
- [x] Docs have been updated

- [x] Feature

Cache compression is mandatory

**What is the new behavior?**
One can opt out from compressing cache using the new `cacheCompression: boolean` option. 
By default set to `true` so it doesn't break anything.

- [x] No

* Impact: couple % faster builds for large projects (1k+ transpiled files)
* Github Issue(s) this is regarding: Resolves #571 

**Other information**:
I wrote a bit about in related issue #571

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-> This readme is for babel-loader v8 + Babel v7  
+> This readme is for babel-loader v8 + Babel v7
 > Check the [7.x branch](https://github.com/babel/babel-loader/tree/7.x) for docs with Babel v6
 
 [![NPM Status](https://img.shields.io/npm/v/babel-loader.svg?style=flat)](https://www.npmjs.com/package/babel-loader)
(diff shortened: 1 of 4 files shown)


danez:
The question we still have is if we maybe remove the cache alltogether and instead recommend using cache-loader.

Have you tried using cache-loader in front of babel-loader that should be faster than the babel cache.

thymikee:
@danez, I didn't try it yet, but I'll try to got to that soon. I've heard Webpack 5 is going to focus on better build cache system – I wonder what are the ideas for caching transformers like Babel or TS with sourcemaps, also for `cache-loader`.

BTW, I have another branch locally where I split code from maps and save both to FS (if necessary) – this way only maps need to be serialized, code goes straight to the disk. The wins are only visible for inline source maps (~5-10% IIRC) though, otherwise it performs similar, so I'm not so happy about that. 

I'll yet have to profile the code properly using `--inspect` to be sure what's slowing things down instead of guessing 😅.

loganfsmyth:
I rebased your branch over our new async-function implementation, and made 2 small changes:

* Use the name `cacheCompression` in the cache function options too
* Move the filename `.gz`-injection logic to the `read`/`write` functions.

loganfsmyth:
Thanks!

PR #583, opened 2018-02-25
Update to latest babel version and fix babelrc behaviour

danez:
This removes the custom behaviour in babel-loader of the option
`babelrc`. It now cannot be a string anymore and should be boolean and instead
`extends` should be used.

BREAKING CHANGE: Option `babelrc` needs to be boolean now. String values are not allowed anymore. Use `extends` option to specify a specific config file. A warning will be triggered when setting babelrc to a string.

### DIFF
--- a/test/cache.test.js
+++ b/test/cache.test.js
@@ -16,7 +16,7 @@ const babelLoader = path.join(__dirname, "../lib");
 const globalConfig = {
   entry: path.join(__dirname, "fixtures/basic.js"),
   module: {
-    loaders: [
+    rules: [
       {
         test: /\.js$/,
         loader: babelLoader,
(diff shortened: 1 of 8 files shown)


danez:
@loganfsmyth It would be nice if babel-core could somehow expose which config files are used for a specific source file. That way we wouldn't have to do the resolve in the loader.

test/cache.test.js
```diff
@@ -288,7 +300,8 @@ test.cb("should allow to specify the .babelrc file", t => {
             exclude: /node_modules/,
             query: {
               cacheDirectory: t.context.cacheDirectory,
-              babelrc: path.join(__dirname, "fixtures/babelrc"),
+              extends: path.join(__dirname, "fixtures/babelrc"),
+              babelrc: false,
```
danez: This is the breaking change basically, I couldn#t figure out why we used a string value for `babelrc` in the first place.

PR #617, opened 2018-05-03
use modules false to enable webpack tree shaking

pravdomil:
by default all modules are transpiled to `commonjs` therefore they cannot be statically analysed and tree shaking will not work

see https://stackoverflow.com/questions/45384170/how-to-fix-modules-with-moduleconcatenation-bailout-module-is-not-an-ecmascrip

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,9 @@ module: {
         loader: 'babel-loader',
         options: {
           presets: [['@babel/preset-env', { modules: false }]]
+          // you need set modules: false in order to get tree shaking working
+          // by default preset-env transpiles modules to commonjs but webpack supports ES6 modules
+          // see https://new.babeljs.io/docs/en/next/babel-preset-env.html#modules
         }
       }
     }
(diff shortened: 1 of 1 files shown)


xtuc:
Could you please add a note in the README to clarify why we need it?

Also that won't work a certain version of Webpack.

pravdomil:
updated

pravdomil:
https://stackoverflow.com/questions/45384170/how-to-fix-modules-with-moduleconcatenation-bailout-module-is-not-an-ecmascrip

pravdomil:
@xtuc can you take a look at it?

xtuc:
Yes that better, could you please indicate which version of Webpack is required for `modules: false`?

pravdomil:
I think that webpack supports es6 modules from the beginning?

Qix-:
This doesn't fix anything. I'm still getting `ModuleConcatenation bailout: Module is not an ECMAScript module`.

loganfsmyth:
This should be all set with #660 now.

PR #629, opened 2018-06-22
Return full result from `babel.transform`.

izaakschroeder:
Previous code only returns an object of the shape `{ code, map, metadata }` which prevents accessing things like the AST when `babel` is configured to output it. This change allows access to more things that `babel` spits out.

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Returns only `{ code, map, metadata }`.

**What is the new behavior?**

Returns more result data (`ast` and `sourceType`).

- [ ] Yes
- [x] No

### DIFF
--- a/src/transform.js
+++ b/src/transform.js
@@ -14,13 +14,18 @@ module.exports = async function(source, options) {
 
   if (!result) return null;
 
-  const { code, map, metadata } = result;
+  // We don't return the full result here because some entries are not
+  // really serializable. For a full list of properties see here:
+  // https://github.com/babel/babel/blob/master/packages/babel-core/src/transformation/index.js
+  // For discussion on this topic see here:
+  // https://github.com/babel/babel-loader/pull/629
+  const { ast, code, map, metadata, sourceType } = result;
 
   if (map && (!map.sourcesContent || !map.sourcesContent.length)) {
     map.sourcesContent = [source];
   }
 
-  return { code, map, metadata };
+  return { ast, code, map, metadata, sourceType };
 };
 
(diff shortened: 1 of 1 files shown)


src/transform.js
```diff
@@ -14,13 +14,13 @@ module.exports = async function(source, options) {
...
 
-  const { code, map, metadata } = result;
+  const { map } = result;
 
   if (map && (!map.sourcesContent || !map.sourcesContent.length)) {
     map.sourcesContent = [source];
   }
 
-  return { code, map, metadata };
+  return result;
```
loganfsmyth: My one concern here is that `result.options` is not serializable to JSON. This change makes the full `options` object available to `result` override callbacks, but it'll only be useful if you're running without the cache, since the cache version would reload and be missing stuff.

When you asked originally I was fine with removing the whitelist of props, but now I'm leaning toward saying we'd be better to at least blacklist the `result.options` property.

izaakschroeder: Good point. I can alternatively just add `ast` to the whitelist if that's better?

loganfsmyth: The full set of props is at https://github.com/babel/babel/blob/9283efaeb7a8c5b505b19332c3407dc44c668777/packages/babel-core/src/transformation/index.js#L64

If you want to expand the whitelist to include all of those except `options` that sounds good to me.

izaakschroeder: Will do! 🎉

izaakschroeder:
All set @loganfsmyth 😄

loganfsmyth:
Thanks!

PR #637, opened 2018-07-05
Avoid absolute paths in cache filename hashes

### LINKED ISSUE #634: Prune babel cache?

Would it be possible to add a way to prune the babel-loader cache after webpack runs?  We've started uploading/downloading our cache in CI to help performance across builds, but the babel-cache grows unbounded and eventually bogs down our CI.

**Babel Loader Version**:
7.1.4

**Current behavior:**
babel cache grows unbounded

It doesn't.

lencioni:
By default, options will look something like:

```js
{
  inputSourceMap: undefined,
  sourceRoot: '/abs/path/to/project',
  filename: '/abs/path/to/project/src/index.js',
  sourceMap: false,
  sourceFileName: 'src/index.js',
}
```

The absolute paths here are problematic when running this on production
build machines that persist this cache but also have different absolute
paths. In this scenario, this causes the cache to be useless, and to
grow in size exceptionally fast.

Since the source filename has enough unique information here about
the cached file, and it is a relative version of the filename, we can
improve this by removing the absolute paths from the options before
hashing.

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

At Airbnb, we are using babel-loader with the cacheDirectory option turned on. For our production builds, we persist this cache across builds. However, each build is run on a different machine in a random temp directory. And, since the absolute paths are currently part of the cache filename hash, this causes the cache to be unused and leads to problems with the cache growing in size too quickly.

**What is the new behavior?**

In this PR, I am removing the absolute paths from the options object before stringifying and hashing.

- [ ] Yes
- [x] No

**Other information**:

I think this should effectively resolve #634.

I wrote this PR against the 7.x branch in hopes that we could get it into a 7.x release. If you think this is a good approach, I am happy to open a similar PR against master.

Also, I didn't see an easy way to write a test for this, so if you have any suggestions on that front, please let me know!

### DIFF
--- a/src/fs-cache.js
+++ b/src/fs-cache.js
@@ -72,9 +72,31 @@ const write = function(filename, result, callback) {
  */
 const filename = function(source, identifier, options) {
   const hash = crypto.createHash("md4");
+
+  // By default, options will look something like:
+  //
+  // {
+  //   inputSourceMap: undefined,
+  //   sourceRoot: '/abs/path/to/project',
+  //   filename: '/abs/path/to/project/src/index.js',
+  //   sourceMap: false,
+  //   sourceFileName: 'src/index.js',
+  // }
+  //
+  // The absolute paths here are problematic when running this on production
+  // build machines that persist this cache but also have different absolute
+  // paths. Since the source filename has enough unique information here about
+  // the cached file, and it is a relative version of the filename, we can
+  // improve this by removing the absolute paths from the options before
+  // hashing.
+  const processedOptions = Object.assign({}, options, {
+    sourceRoot: null,
+    filename: null,
+  });
+
   const contents = JSON.stringify({
     source: source,
-    options: options,
+    options: processedOptions,
     identifier: identifier,
   });
 


loganfsmyth:
My concern with this is that it's possible that some user's plugin would be using the whole file path, so by doing this you'd risk a cache collision. That immediately means I'm against it being any kind of default behavior.

I think I'd be alright with having this be an option, if you'd like, where you could pass in the root folder of the project, and we'd strip that specific prefix off of absolute paths.

I will say, it still seems a bit like a hack, since there's nothing special about Babel's approach here, and any build tooling anywhere in your process could have similar issues. You'd probably be better off trying to make your build environment consistent by using `chroot` or containers.

lencioni:
Thanks for the quick review on this PR! 

> some user's plugin would be using the whole file path, so by doing this you'd risk a cache collision

By "some user's plugin" you are referring to a babel plugin, right? Although that seems unlikely, yes, that's a good point that would lead to a cache collision!

> I think I'd be alright with having this be an option, if you'd like, where you could pass in the root folder of the project, and we'd strip that specific prefix off of absolute paths.

Yeah I think that would work for us!

> I will say, it still seems a bit like a hack, since there's nothing special about Babel's approach here, and any build tooling anywhere in your process could have similar issues. You'd probably be better off trying to make your build environment consistent by using `chroot` or containers.

I think you are correct, and this is an issue that we have addressed elsewhere before. I'll look into `chroot` and if I can't use that for some reason, I'll update this PR with the option approach.

With that said, however, I ran another test on our build machine that simply ran webpack twice in a row in the same job, and the cache didn't seem to speed it up at all (10-11 minutes per run). So, if the cache isn't helping improve speeds for us right now, it seems that even fixing these cache keys won't give us the speed boost we were hoping for. I think I need to investigate that first as well. Let me know if you happen to have any suggestions for things to look for or try!

lencioni:
It sounds like we are moving our CI stuff to a containerized system, so once we are able to fully migrate to that, hopefully this problem will disappear for us. I'm going to close this PR for now, thanks again for the speedy review!

PR #638, opened 2018-07-05
Use md4 instead of sha1 for filename hashes

lencioni:
We don't need the cryptographic strength of SHA-1 for this purpose, so
we may as well use a faster hashing algorithm.

While I was at it, I also sapped out the use of hash.end + hash.read
combo for the faster and more conventional hash.update + hash.digest. In
my local testing, this also seems to offer a nice speed boost.

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Uses SHA-1 for cache filename hash.

**What is the new behavior?**

Uses MD4 for cache filename hash.

- [ ] Yes
- [x] No

**Other information**:

I wrote this PR against the 7.x branch in hopes that we could get it into a 7.x release. If you think this is a good approach, I am happy to open a similar PR against master.

### DIFF
--- a/src/fs-cache.js
+++ b/src/fs-cache.js
@@ -71,16 +71,16 @@ const write = function(filename, result, callback) {
  * @return {String}
  */
 const filename = function(source, identifier, options) {
-  const hash = crypto.createHash("SHA1");
+  const hash = crypto.createHash("md4");
   const contents = JSON.stringify({
     source: source,
     options: options,
     identifier: identifier,
   });
 
-  hash.end(contents);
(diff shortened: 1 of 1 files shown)


existentialism:
We should land this change in `master` too?

xtuc:
Yes that's a good idea.

hzoo:
And we can release this as a patch?

hzoo:
https://github.com/babel/babel-loader/releases/tag/v7.1.5

PR #646, opened 2018-07-30
Resolve presets if resolveLoader.modules is set (7.x)

johanflint:
If `resolveLoader.modules` is specified, it is ignored for presets (see #166 ). This fixes the problem by resolving the presets first. 

I'm not 100% sure this is the way it should be fixed, but this fix works for our React Native project where all our dependencies and loaders are not in the same directory as the sources. We're still running React Native 55.x, which uses Babel 6, hence the fix on version 7.x. If the patch is accepted, we'll open one for 8.x (as we're in the process of upgrading React Native and Babel).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -148,6 +148,14 @@ module.exports = function(source, inputSourceMap) {
     options.sourceFileName = relative(options.sourceRoot, options.filename);
   }
 
+  if (options.presets !== undefined) {
+    options.presets = loaderOptions.presets.map(function(preset) {
+      return typeof preset === "string"
+        ? require.resolve(`babel-preset-${preset}`)
+        : [require.resolve(`babel-preset-${preset[0]}`), preset[1]];
+    });
+  }
+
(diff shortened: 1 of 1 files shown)


loganfsmyth:
You can resolve the paths to the plugins/presets yourself before passing them to `babel-loader`, but this isn't something we'd want to land in `babel-loader` itself. Babel already has rules around how it resolves plugins/presets, and special-casing that specifically for babel-loader would complicate things and add to the confusion.

> If resolveLoader.modules is specified, it is ignored for presets

That value does not affect Babel's own internal rules for resolving plugins. It is not ignored, it just does not do what you'd like it to.

johanflint:
Allright, that makes sense. It took us some more debugging to find out that [haul](https://github.com/callstack/haul) added the preset if no `.babelrc` file is present and haul doesn't resolve it before passing it to babel. I can work around it there.

Thank you for taking your time to looking at this pull request. 👍

PR #650, opened 2018-08-07
Add `modules: false` by default for es2015/env/latest presets

### LINKED ISSUE #477: Warning if transpiling ES modules to commonjs

### LINKED ISSUE #521: Set `modules: false` by default on babel-preset-env

SpaceK33z:
**Note** this is an update version of PR #529.

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

See #521.

**What is the new behavior?**

The Babel presets `env`, `latest` and `es2015` will be mutated to add `modules: false` as an option. This also works when you use e.g. `@babel/preset-env` or `@babel/env`.

- [x] Yes
- [ ] No

For people who use babel-loader without setting the `modules` option, there might be small differences in the way webpack handles ES modules.

For people who still want to keep the old behavior, they can explicitly add `modules: true` or e.g. `modules: "amd"`.

**Other information**:

I am not happy at all with the implementation, but it works. Let me know if there is a better way I missed.

I used parts of the code in #485, so credits to @wardpeet!

Fixes #521, fixes #477, fixes #485

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -85,6 +85,28 @@ async function loader(source, inputSourceMap, overrides) {
     );
   }
 
+  if (this.version > 1 && programmaticOptions.presets) {
+    programmaticOptions.presets.forEach((preset, i) => {
+      let name = preset;
+      let opts = {};
+      if (Array.isArray(preset)) {
+        name = preset[0];
+        opts = preset[1];
+      }
+      if (
+        ["env", "babel-preset-env", "@babel/preset-env"].includes(name) &&
+        (!opts || (opts && !opts.hasOwnProperty("modules")))
+      ) {
+        const newOpts = { modules: false };
+        if (Array.isArray(preset)) {
+          preset[1] = Object.assign({}, opts, newOpts);
+        } else {
+          programmaticOptions.presets[i] = [name, newOpts];
+        }
+      }
+    });
+  }
+
   const config = babel.loadPartialConfig(programmaticOptions);
   if (config) {
     let options = config.options;
--- a/src/index.js
+++ b/src/index.js
@@ -87,18 +87,15 @@ async function loader(source, inputSourceMap, overrides) {
 
   if (this.version > 1 && programmaticOptions.presets) {
     programmaticOptions.presets.forEach((preset, i) => {
-      let name = preset;
-      let opts = {};
-      if (Array.isArray(preset)) {
-        name = preset[0];
-        opts = preset[1];
-      }
+      const presetIsArray = Array.isArray(preset);
+      const name = presetIsArray ? preset[0] : preset;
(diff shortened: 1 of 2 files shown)


Andarist:
I think there should be something done, because current situation is way from ideal and people shouldn't worry about it (unless they want to). But change such as this doesnt account for `.babelrc`/`.babelrc.js`/`package.json#babel`/`babel.config.js` lookup.

SpaceK33z:
@Andarist, you're right, it doesn't work for `.babelrc` etc right now so this is kinda useless. Didn't notice that.

Spent some time trying to fix this, but got stuck at not being able to mutate the [`options` from a `ConfigItem`](https://github.com/babel/babel/blob/b6db036ff77993cf8e781c05380b36bdba75fa89/packages/babel-core/src/config/item.js#L76). I'm completely new to the Babel codebase, so would you have any pointers on how to easily do that?

Andarist:
>  I'm completely new to the Babel codebase, so would you have any pointers on how to easily do that?

I'm not familiar with config loading myself. In general I think it would be better to validate the config rather than mutate it, but also ain't sure how to do it right now. @loganfsmyth should know if it's even possible right now, or if there is anything that can be done in babel itself to ease this.

Andarist:
Seems like this should be closed in favour of https://github.com/babel/babel-loader/pull/660 . Thanks for making this PR - it certainly pushed the needed changes forward so that this can get resolved once and for all now :)

PR #655, opened 2018-08-13
Added option for custom cache hash behavior

Saionaro:

- [ ] Bugfix
- [x] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

No option for setting custom cache hash behavior

**What is the new behavior?**

Added option for setting custom cache hash behavior.
```
const fields = {
  filename: 1,
  sourceRoot: 1,
  dirname: 1,
  resolved: 1,
  cwd: 1
};
...
modifyCacheHash: (source, options, cacheIdentifier) => {
  const convertedOptions = JSON.stringify(options, (key, value) => {
    if (fields[key]) {
      return undefined;
    }
    return value;
  });
  return source + convertedOptions + cacheIdentifier;
}
...
```

- [ ] Yes
- [x] No

**Other information:**

It can be useful if your builds are deploying on Heroku and a build directory is dynamic (for example `/tmp/build<random-hash>/`). In this case, we should exclude the absolute build path from the hash string to hit cache stored by Heroku.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,13 @@ This loader also supports the following loader-specific option:
 
 * `cacheIdentifier`: Default is a string composed by the babel-core's version, the babel-loader's version, the contents of .babelrc file if it exists and the value of the environment variable `BABEL_ENV` with a fallback to the `NODE_ENV` environment variable. This can be set to a custom value to force cache busting if the identifier changes.
 
+* `modifyCacheHash`: Default `undefined`. When setted, will invokes as  
+`function(source, options, cacheIdentifier) -> Object|String` where  
+`String source` - module's code;  
+`Object options` - united options of loader;  
+`String cacheIdentifier` - see above.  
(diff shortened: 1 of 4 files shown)


Saionaro:
Hello, @hzoo ! Please, take a look - WDYT?

edmorley:
Indeed, with Heroku (or any other deployment or CI service that both varies the build directory per job, and allows for persisting a cache directory), without this feature the cache is useless.

@lencioni opened #637 to handle this in a different way, so it seems like there is demand. (And I suspect that many people don't realise their caches are being busted, and just think the cache feature isn't that effective.)

Saionaro:
@loganfsmyth can you take a look at it?

loganfsmyth:
I think my feeling here is the same as https://github.com/babel/babel-loader/pull/637#issuecomment-402767160 We could add a `cacheRoot` or something if you want, and automatically strip that out. There's still always the risk that it'll go wrong though. I really seems like something Heroku should be resolving using chroot or something so the builds run in a consistent directory location.

loganfsmyth:
Got some useful feedback from someone at Heroku in https://github.com/heroku/heroku-buildpack-nodejs/issues/439#issuecomment-419557813 FYI. If it's not going to happen on their side any time soon, I'm definitely more willing to add a `cacheRoot` option. Still not a huge fan a generic `modifyCacheHash` function though.

PR #670, opened 2018-09-02
Use 'sourceMaps' since that is what we suggest to use in our docs.

### LINKED ISSUE #658: Use of sourceMap alias of sourceMaps.
babel-loader uses the `sourceMaps` option alias of `sourceMap` which will cause a validation fail if a user tries to use the documented `sourceMaps` option (instead of the `sourceMap` alias).

loganfsmyth:
Fixes #658

Babel allows both 'sourceMap' and 'sourceMaps'. Not only were we using the wrong one (because our docs recommend 'sourceMaps', but we were also causing Babel to _always_ throw an unhelpful error when the user used the wrong one.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -75,16 +75,29 @@ async function loader(source, inputSourceMap, overrides) {
     );
   }
 
+  // Standardize on 'sourceMaps' as the key passed through to Webpack, so that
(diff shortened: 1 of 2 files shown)


jdalton:
Thank you @loganfsmyth!

PR #672, opened 2018-09-03
Use .catch instead of catching in .then

vikr01:

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

A `.then` block after a Promise was checking for the return value and for errors.

**What is the new behavior?**
Use a `.catch` block after the Promise to check for errors instead of checking for errors in the `.then`.

- [ ] Yes
- [x] No

* Impact:

**Other information**:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -43,7 +43,8 @@ function makeLoader(callback) {
 
     loader
       .call(this, source, inputSourceMap, overrides)
-      .then(args => callback(null, ...args), err => callback(err));
+      .then(args => callback(null, ...args))
+      .catch(err => callback(err));
   };
 }
 


src/index.js
```diff
@@ -43,7 +43,8 @@ function makeLoader(callback) {
 
     loader
       .call(this, source, inputSourceMap, overrides)
-      .then(args => callback(null, ...args), err => callback(err));
+      .then(args => callback(null, ...args))
+      .catch(err => callback(err));
```
loganfsmyth: While theoretically pretty similar, this is technically worse because it means that if `callback(null, ...args)` threw an exception, it would call `callback(err)`, meaning `callback` would get called with an error that it actually threw itselff, instead of triggering an uncaught error.

PR #685, opened 2018-09-22
Also pass the caller option to loadPartialConfig

hedgepigdaniel:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The caller option (added in https://github.com/babel/babel-loader/pull/660) is passed to the transform function here:

https://github.com/babel/babel-loader/blob/8f240b498bb24ef89f7b306f5ac806e84b813b0d/src/transform.js#L10-L10

But it is not passed to the loadPartialConfig function here:

https://github.com/babel/babel-loader/blob/8f240b498bb24ef89f7b306f5ac806e84b813b0d/src/index.js#L121-L121

This results in the caller option not being available in the project's `babel.config.js`

**What is the new behavior?**

The `caller` option is passed (subject to the existing compatibility check) in both cases.

- [ ] Yes
- [x] No

* Impact:

**Other information**:

I'm not sure why it works for `@babel/preset-env` - I assume the presets referenced from `babel.config.js` get access to the option from the `transform` call rather than the `loadPartialConfig` call.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -22,6 +22,7 @@ if (/^6\./.test(babel.version)) {
 const pkg = require("../package.json");
 const cache = require("./cache");
 const transform = require("./transform");
+const injectCaller = require("./injectCaller");
 
 const loaderUtils = require("loader-utils");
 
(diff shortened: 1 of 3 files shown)


loganfsmyth:
Ah great point. We actually don't need to pass it to `transform` at all in this case, so we can remove it from the `transform` file entirely.

hedgepigdaniel:
Ah, yes that makes sense. I've put the `injectCaller` call in only one place now, before `loadPartialConfig`.

loganfsmyth:
Thanks!

PR #687, opened 2018-09-24
Add `customize` option

Timer:
Hey there!

We're looking to use [`babel-plugin-macros`](https://github.com/kentcdodds/babel-plugin-macros) without disabling caching.
Right now, there are larger talks about how to solve this (see https://github.com/babel/babel/issues/8497) via letting plugins hint to external file dependencies.

In the interim, we figured we'd disable caching *only* for files that include `.macro` or `/macro` in their source.
It would be nice to  [customize the loader](https://github.com/babel/babel-loader#customized-loader) through options instead of creating our own loader.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,8 @@ This loader also supports the following loader-specific option:
 
 * `cacheCompression`: Default `true`. When set, each Babel transform output will be compressed with Gzip. If you want to opt-out of cache compression, set it to `false` -- your project may benefit from this if it transpiles thousands of files.
 
+* `overrides`: Default `null`. See [Customized Loader](#customized-loader).
+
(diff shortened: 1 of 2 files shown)


loganfsmyth:
Alright, landed in https://github.com/babel/babel-loader/commit/9b70a02b3069000133122c09271ce0a7144f8dae but tweaked a bit in https://github.com/babel/babel-loader/commit/ac0c869b86b98881ac5544bf7248c63a971b7219

Timer:
Thanks much for this!

loganfsmyth:
I'm hesitant to publish right before I go to bed in case something screws up, but I'll plan to publish when I get up.

loganfsmyth:
Published `8.0.3`

Timer:
Excellent, thanks!

PR #714, opened 2018-10-29
wip: include dynamic import automatically

### LINKED ISSUE #515: [feature request] include dynamic import syntax plugin automatically
I've written/talked about this before a few times, might as well make it an issue.

We shouldn't have to ask users to include the dynamic import plugin if webpack is supporting it and is a source of pain. Ideally it would work with preset-env where webpack is it's own target or something but this is already babel-loader so we know that a user is using webpack + babel together. We could just make it a dependency (or just use `parserOpts: { plugins: ['dynamicImport'] }` and no dependency necessary).

So like at https://github.com/babel/babel-loader/blob/1ca5c7856d2129a28d6ba031bd8a5759aa80b8eb/src/index.js#L128-L129

```diff
const defaultOptions = {
+  parserOpts: { plugins: ['dynamicImport'] },
```

Or we'd have to do append the plugin to the end of the `plugins`: 

```js
options.plugins.push(require('babel-plugin-syntax-dynamic-import'))
```

thiagoarrais:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

This fixes #515 . The dynamic import syntax plugin has to be manually added to webpack config even though webpack itself already supports dynamic imports.

**What is the new behavior?**

Dynamic import syntax is automatically supported.

- [ ] Yes
- [x] No

**Other information**:

Please let me know if and where I should add docs/tests.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -121,7 +121,8 @@ async function loader(source, inputSourceMap, overrides) {
     delete loaderOptions.sourceMap;
   }
 
-  const programmaticOptions = Object.assign({}, loaderOptions, {
+  const defaultOptions = { parserOpts: { plugins: ["dynamicImport"] } };
+  const programmaticOptions = Object.assign({}, defaultOptions, loaderOptions, {
     filename,
     inputSourceMap: inputSourceMap || undefined,
 


nicolo-ribaudo:
I thibk that we should enable it based on `caller` (e.g. `caller.supportsDynamicImport`).
Also, probably we should only enable it inside `preset-env` since it's the only package which enables anything automatically?

thiagoarrais:
@nicolo-ribaudo: I'm not familiar with this part of the code yet and may be misreading you here. Are you suggesting to move this change to `@babel/preset-env`?

nicolo-ribaudo:
Yes, but I'm not 100% sure about it so lets wait for someone else's opinion.

> I'd prefer that we enable it based on caller (e.g. caller.supportsDynamicImport), like we do for supportESModules).

I think that we should do this in any case, though.

hzoo:
Yes we introduced that flag to add this functionality, it's from https://github.com/babel/babel-loader/pull/660

thiagoarrais:
Am I right to understand that https://github.com/babel/babel-loader/blob/7d8500c/src/injectCaller.js#L13 means that `babel-loader` already flags that dynamic imports are supported? And that we need to change `preset-env` to interpret this flag? If so, I'll close this PR and open another to `preset-env`.

existentialism:
@thiagoarrais yeah you're correct. I actually have a branch with the behavior enabled, but need to dig back into my notes on something I ran into (or some decision that needed to be made)

loganfsmyth:
I added `supportsDynamicImport` on the off chance we wanted it, but I think it's not obvious if we'll want to use it or what.

The main issue in my mind is that it seems surprising to automatically parse this syntax only when running in Webpack and not other contexts when it isn't standard yet. Once we get to the point of it being opt-in, I'm not sure it is clear why we'd go that route instead of expecting people to enable the syntax plugin manually.

thiagoarrais:
I will go on and try to implement this in preset-env as a learning exercise. We can later decide on merging it or not.

@existentialism: if you happen to find that branch, please let me know.

@loganfsmyth: I agree that it is surprising. But as I understand, that was exactly the intended behavior described in the original issue (#515). The project understanding can change, of course. Maybe the OP (@hzoo) can weigh in here?

Closing the PR for now to signal that I won't work on this change any longer. Please do keep the discussion going.

PR #738, opened 2018-11-22
New [transformFileName] loader-specific option has been implemented.

alexandrklimov:
REASON

A reason of implementing this feature is [VueJS TypeScript SFC]
compilation trouble with the usage of [babel, babel-loader].
Issue environment:  webpback@4.25.x,  vue@2.5.x., babel@7.1.x,
babel-loader@8.0.x)

Ordinal *.ts files are compiled OK
but SFC - the [babel-loader] doesn’t understand that
[vue-loader] gives it TypeScript-code.

I know, that similar problem solved for the [ts-loader] by its option
[appendTsSuffixTo], but I haven't found any similar settings for
[babel-loader].

    DESCRIPTION

babel-loader have two file names field in a loader's options object:
 - filename
 - sourceFilename

If I've found out correctly, the last one is strongly bound
to a physical resource (a file) and the first one is a logical name
in some sense.
Consequently, I change the fist file name options.
Well, in my test build that solution works without any negative impact.

I think that the [ts-loader] solves such problem by similar solution.

    OPTION'S STRUCTURE

transformFileName:{
    pattern: RegEx,
    replace: 'string_of_content_instead_of_regex'
}

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "@babel/cli": "^7.0.0",
     "@babel/core": "^7.0.0",
     "@babel/preset-env": "^7.0.0",
+    "@babel/preset-typescript": "^7.1.0",
     "ava": "0.25.0",
     "babel-eslint": "^8.0.0",
     "babel-plugin-istanbul": "^4.0.0",
--- a/src/index.js
+++ b/src/index.js
@@ -121,8 +121,18 @@ async function loader(source, inputSourceMap, overrides) {
     delete loaderOptions.sourceMap;
   }
 
+  let transformedFileName = filename;
+  if (loaderOptions.transformFileName) {
+    const pattern = loaderOptions.transformFileName.pattern;
+    const replace = loaderOptions.transformFileName.replace;
+
+    if (pattern && replace) {
+      transformedFileName = filename.replace(pattern, replace);
+    }
+  }
+
   const programmaticOptions = Object.assign({}, loaderOptions, {
-    filename,
+    filename: transformedFileName,
     inputSourceMap: inputSourceMap || undefined,
 
     // Set the default sourcemap behavior based on Webpack's mapping flag,
@@ -137,7 +147,9 @@ async function loader(source, inputSourceMap, overrides) {
     // modules.
     sourceFileName: filename,
   });
+
   // Remove loader related options
+  delete programmaticOptions.transformFileName;
   delete programmaticOptions.customize;
   delete programmaticOptions.cacheDirectory;
   delete programmaticOptions.cacheIdentifier;
--- a/test/fixtures/TestTS.ts
+++ b/test/fixtures/TestTS.ts
@@ -1,0 +1,12 @@
+export default class TestTsCls{
+  private name: string;
+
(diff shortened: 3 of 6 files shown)


danez:
Thanks for your PR. 
I looked at ts-loader to see how the solve it. I think for the babel loader a more generic option makes sense in comparison to the `appendTs(x)SuffixTo` options.

Why did you choose to have an object with pattern and replacement? I would rather allow a function to be passed `transformFileName: (currentFileName) => newFileName`.

alexandrklimov:
Hello @danez !

Thank you for your review.

I've chosen the RegExp-based approach instead of function-base one coz I thought just about a context of my task - file name transdormation for right compilation.
Moreover, I am not too professional in JS development and in Babel using particularly and I try to provide a solution is as concise as possible.

Of course, your approach with a mapper function is more generic and flexible.

asselinpaul:
@alexandrklimov 
Hi Alex, 
Are you aware of a way to get babel-loader to understand that vue-loader gives it Typescript code?
Best —
Paul

alexandrklimov:
Hi @asselinpaul 

As far I know, if you don't use **.vue** files, **babel-loader** should be configured as usual.

asselinpaul:
Forgot to mention, I am using **.vue** files 😟 . It works fine for **.ts**. @alexandrklimov

alexandrklimov:
Unfortunately there isn't any way to ask Babel to detect **.vue**-files as **.ts** ones, through **babel-loader**.
Or I just don't known about this way :)

This pull-request have been made for resolving this issue in **tsc**-style [manner ](https://github.com/vuejs/vetur/issues/213#issuecomment-303020014) but it wasn't been applied at the first time.
Then I closed it accidentally and then after reopening I got current conflicts.

Though, I can fix these conflicts but there isn't any guarantee that the repository owner will apply my solution.

alexandrklimov:
It's so sad but I can't sight any progress in resolving this annoying problem with any way solution.

asselinpaul:
@alexandrklimov let's see if this gets anywhere — I do see @danez 's point of using a function instead of the regex.

alexandrklimov:
@asselinpaul Yes, I remember about this suggestion and it's absolutely OK for me, but I couldn't find neither any remark about that solution in documentation neither any commit, since the last my converation with @danez.
Be frankly, I check this problem the last time about a month ago.

Suggestion about function I treated just as a suggestion - may be it was a hint? :)

If a function is a final repository owner's solution - it's OK for me. I can fix my fix for it.
I just would like to know a direction to appropriated pull request.

asselinpaul:
This is still relevant to me. Happy to help if my help is required.

asselinpaul:
https://github.com/Realytics/fork-ts-checker-webpack-plugin let's me typecheck TypeScript inside of Vue SFC (.vue) files but babel-loader still (rightfully) freaks out at the TypeScript syntax.

light0x00:
i have same problem...

Jdruwe:
Any update on this, I would also like to use the babel-loader for .vue files.

PR #750, opened 2018-12-23
Remove docs mention about sourceMap option being ignored

Andarist:
If I'm not mistaken when reading the source code - this note is not true (maybe it once was true, not sure)


loganfsmyth:
Thanks!

Andarist:
OTOH maybe the documented behaviour should actually be implemented? I dont find a use case when somebody would like to i.e. have webpack sourcemaps enabled but babel ones disabled 🤔 (and vice versa ofc)

PR #773, opened 2019-03-01
Fix babel stackoverflow link

alexander-schranz:
The issue template is linking to the false stackoverflow tag

http://stackoverflow.com/questions/tagged/babel

should be replaced with:

http://stackoverflow.com/questions/tagged/babeljs


danez:
Thanks

Issue #787, opened 2019-05-09
Provide plugins with insight into webpack target


DylanPiercey:
**I'm submitting a feature request**
I am currently using server side and client side bundling with webpack and babel-loader. I believe this pattern is going to increase since it improves performance (especially with lambda's), and can even simplify the toolchain developers use (shared config between server and browser).

One small change I think could help simplify basic configs though would be to make it so that the webpack `target` was passed into the babel `caller` option.

Currently if I wan't to use a `.babelrc` or `babel.config.js` I have to pass this information myself so that I can efficiently use `@babel/preset-env` with `{ node: "current" }` for the server and just the defaults in the browser.

To facilitate this my webpack config becomes something like:

```js
const serverConfig = {
  ...sharedConfig,
  module: {
    rules: [
      {
        test: /\.js$/,
        use: [
          {
            loader: "babel-loader",
            options: {
              caller: {
                target: "node"
              }
            }
          }
        ]
      },
      ...
    ]
  }
}

const browserConfig = {
  ...sharedConfig,
  module: {
    rules: [
      {
        test: /\.js$/,
        use: [
          {
            loader: "babel-loader",
            options: {
              caller: {
                target: "web"
              }
            }
          }
        ]
      },
      ...
    ]
  }
}
```

This is a lot of config that I can't share just so that I can compile my files with babel differently depending on environment.

My `babel.config.js` would look something like:

```js
module.exports = api => {
  return {
    presets: [
      [
        "@babel/env",
        {
          loose: true,
          targets: api.caller(isNode) ? {
            node: "current"
          } : undefined
        }
      ]
    ]
  };
};

function isNode(caller) {
  return caller.target === "node";
}
```

If the `target` was made available to babel via `caller` I could instead add just the following to my shared config which simplifies it a lot!

```js
const sharedConfig = {
  module: {
    rules: [
      {
        test: /\.js$/,
        loader: "babel-loader"
      },
      ...
    ]
  }
}
```

**Webpack Version:**
4.30.0

**Babel Core Version**:
7.4.4

**Babel Loader Version**:
8.0.5

**Please tell us about your environment:**
OSX 10.x

**Current behavior:**
The current webpack output `target` is not made available to babel plugins.

**Expected/desired behavior:**
Provide the current webpack output `target` via `caller` to babel plugins.

---

Happy to create a PR for this, and perhaps even standardize this across the different babel bundler plugins.

DylanPiercey:
[...]

PR #826, opened 2020-02-27
feat: expose webpack target via babel caller

DylanPiercey:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently it is not possible to determine inside of a `babel.config.js` file what the webpack target being bundled for is.

**What is the new behavior?**
Exposes the current webpack target using the babel [caller](https://babeljs.io/docs/en/config-files#apicallercb) api.

- [ ] Yes
- [x] No

**Other information**:
I'm happy to add a test if necessary. However there are not existing tests for the already injected caller data. I think it'd require a bit of effort to setup correctly since we'd probably want to write a fake plugin which just checked what the caller was.

Implements #787

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,9 +193,9 @@ In the case one of your dependencies is installing `babel` and you cannot uninst
   }
 ```
 
-## Customize config for specific Webpack target
+## Customize config based on webpack target
 
-Webpack supports bundling for multiple different [targets](https://webpack.js.org/concepts/targets/) including the web, and node. Sometimes it can be useful to have a slightly different babel configuration for the different environments. This loader provides a `target` property via the babel [caller](https://babeljs.io/docs/en/config-files#apicallercb) API which can be used in this case.
+Webpack supports bundling multiple [targets](https://webpack.js.org/concepts/targets/). For cases where you may want different Babel configurations for each target (like `web` _and_ `node`), this loader provides a `target` property via Babel's [caller](https://babeljs.io/docs/en/config-files#apicallercb) API.
 
 For example, to change the environment targets passed to `@babel/preset-env` based on the webpack target:
 
--- a/README.md
+++ b/README.md
@@ -193,6 +193,37 @@ In the case one of your dependencies is installing `babel` and you cannot uninst
   }
 ```
 
+## Customize config for specific Webpack target
+
+Webpack supports bundling for multiple different [targets](https://webpack.js.org/concepts/targets/) including the web, and node. Sometimes it can be useful to have a slightly different babel configuration for the different environments. This loader provides a `target` property via the babel [caller](https://babeljs.io/docs/en/config-files#apicallercb) API which can be used in this case.
+
+For example, to change the environment targets passed to `@babel/preset-env` based on the webpack target:
+
+```javascript
+// babel.config.js
+
+module.exports = api => {
+  return {
+    plugins: [
+      "@babel/plugin-proposal-nullish-coalescing-operator",
+      "@babel/plugin-proposal-optional-chaining"
+    ],
+    presets: [
+      [
+        "@babel/preset-env",
+        {
+          useBuiltIns: "entry",
+          // caller.target will be the same as the target option from webpack
+          targets: api.caller(caller => caller && caller.target === "node")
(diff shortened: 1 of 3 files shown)


nicolo-ribaudo:
`target` can also be a function, but I wonder how it interacts with config caching :thinking:

DylanPiercey:
@nicolo-ribaudo hmm. I thought it was resolved to a string by the time it got to the loader. Could very well be wrong on that though. I'm pretty sure `caller` will error if passed a function since they serialize it for the cache key.

nicolo-ribaudo:
> I thought it was resolved to a string by the time it got to the loader.

Oh ok, I didn't know about it :+1:

DylanPiercey:
Yes it looks like it is resolved to a string since the `target` for the loader is actually added by a plugin here: https://github.com/webpack/webpack/blob/0d4607c68e04a659fa58499e1332c97d5376368a/lib/LoaderTargetPlugin.js#L29 which is always passed a string. In the function target example in the webpack docs they actually use this plugin inside the function (https://webpack.js.org/configuration/target/#function).

README.md
```diff
@@ -193,6 +193,38 @@ In the case one of your dependencies is installing `babel` and you cannot uninst
   }
 ```
 
+## Customize config for specific Webpack target
+
+Webpack supports bundling for multiple different [targets](https://webpack.js.org/concepts/targets/) including the web, and node. Sometimes it can be useful to have a slightly different babel configuration for the different environments. This loader provides a `target` property via the babel [caller](https://babeljs.io/docs/en/config-files#apicallercb) API which can be used in this case:
```
existentialism:
```suggestion
Webpack supports bundling for multiple [targets](https://webpack.js.org/concepts/targets/). For cases where you may want different Babel configurations for each target (like `web` _and_ `node`), this loader provides a `target` property via Babel's [caller](https://babeljs.io/docs/en/config-files#apicallercb) API:
```

README.md
```diff
@@ -193,6 +193,38 @@ In the case one of your dependencies is installing `babel` and you cannot uninst
...
+      "@babel/plugin-proposal-optional-chaining"
+    ],
+    presets: [
+      [
+        "@babel/preset-env",
+        {
+          useBuiltIns: "entry",
+          // caller.target will be the same as the target option from webpack
+          ...(api.caller(caller => caller.target === "node")
+            ? { node: "current" }
```
existentialism: Shouldn't this also be `{ targets: { node: 'current' }}`?

README.md
```diff
@@ -193,6 +193,38 @@ In the case one of your dependencies is installing `babel` and you cannot uninst
...
+            : { targets: { chrome: "58", ie: "11" } }
+          )
+        }
+      ]
+    ]
+  }
+}
+```
+
+In the above `babel.config.js` we're able to distinguish if we are transpiling for `node` or the `web` and adjust accordingly. In this case it has been configured such that `node` targets use `@babel/preset-env` to transpile features for the current version of node, while the `web` target will transpile features back to ie 11 and Chrome 58.
```
existentialism: I feel like we can skip explaining the targets with a quick description of the example (maybe with something like what I suggested above)?

Thoughts?

README.md
DylanPiercey: Yeup, you're right. I've gone ahead and fixed that up. Thanks!

README.md
DylanPiercey: I agree. I have incorporated these changes in the PR. Thanks for the feedback 😄

existentialism:
@DylanPiercey thanks!

bpinto:
Is this working? I can't get anything other than `web` as my target, even though it's configured to use a browserlist environment `target: 'browserslist:app'`.

bpinto:
@DylanPiercey so there is not way to use webpack `target` (raw) for babel-loader `browserlistEnv` option? Thank you.

DylanPiercey:
@bpinto I think it probably could be exposed as well, but it'd need to be a new property on the caller. Currently it is not exposed.

bpinto:
@DylanPiercey https://github.com/babel/babel-loader/pull/943

PR #841, opened 2020-03-25
Webpack is not a peerDependencies

bertho-zero:
Webpack is not used in sources and babel-loader does not directly depend on it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,7 @@
     "schema-utils": "^2.6.5"
   },
   "peerDependencies": {
-    "@babel/core": "^7.0.0",
-    "webpack": ">=2"
+    "@babel/core": "^7.0.0"
   },
   "devDependencies": {
     "@babel/cli": "^7.2.0",


package.json
```diff
@@ -17,8 +17,7 @@
     "schema-utils": "^2.6.5"
   },
   "peerDependencies": {
-    "@babel/core": "^7.0.0",
-    "webpack": ">=2"
+    "@babel/core": "^7.0.0"
```
JLHwung: I believe specifying `webpack` peerDependencies is a de-facto convention of webpack plugins.

css-loader: https://github.com/webpack-contrib/css-loader/blob/master/package.json#L43
file-loader: https://github.com/webpack-contrib/file-loader/blob/master/package.json#L41
etc.

bertho-zero: I use a storybook that needs `babel-loader`, and storybook uses its `webpack` version.
I don't need a `webpack` in my repo, there is only a babel build there.

Maybe it's storybook to add `webpack` to peerDependencies then?

JLHwung: > Maybe it's storybook to add webpack to peerDependencies then?

Correct.

PR #878, opened 2020-11-09
Use Yarn 2

nicolo-ribaudo:


babel.config.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  ignore: [/\.pnp\.c?js/],
```
nicolo-ribaudo, line 2: This is because AVA was running Babel on .pnp.js

PR #822, opened 2020-02-01
Add schema validation

ogonkov:
**Please Read the [CONTRIBUTING Guidelines](https://github.com/babel/babel-loader/blob/master/CONTRIBUTING.md)**
*In particular the portion on Commit Message Formatting*

**Please check if the PR fulfills these requirements**
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

**What is the current behavior?** (You can also link to an open issue here)

Options is validated programmatically

**What is the new behavior?**

Options validated through schema, webpack will assist if given option is not match schema

**Does this PR introduce a breaking change?**
- [ ] Yes
- [x] No

If this PR contains a breaking change, please describe the following...

* Impact:
* Migration path for existing applications:
* Github Issue(s) this is regarding:

**Other information**:
Closes #749

PR #882, opened 2020-11-10
Test on Node.js 8

nicolo-ribaudo:
https://github.com/babel/babel-loader/blob/a4d17c0a6a0693bf970a65dd0a1cca783f1c03a6/package.json#L10

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,8 @@ jobs:
   test-legacy:
     name: Test - ubuntu-latest - Node v8.9, Webpack 4
     runs-on: ubuntu-latest
+    env:
+      YARN_NODE_LINKER: node-modules
     steps:
       - uses: actions/checkout@v2
       - name: Use Node.js 14.x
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -48,3 +48,32 @@ jobs:
...
+        env:
+          YARN_NODE_LINKER: node-modules
+      - name: Install webpack 4
+        run: yarn add -D webpack@4
+        env:
+          YARN_NODE_LINKER: node-modules
+      - name: Build babel-loader
+        run: yarn run build
+        env:
+          YARN_NODE_LINKER: node-modules
```
JLHwung, line 71: The env `YARN_NODE_LINKER ` can be moved to the job scope: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#env

package.json
```diff
@@ -7,7 +7,7 @@
   ],
   "main": "lib/index.js",
   "engines": {
-    "node": ">= 6.9"
+    "node": ">= 8.9"
```
nicolo-ribaudo, line 10: This was a breaking change in https://github.com/babel/babel-loader/pull/822, but no one complained :shrug: 

Also, AVA doesn't work on Node.js 8 so I'm running a smaller single test.

existentialism: 🤔

nicolo-ribaudo: `schema-utils`, introduced in that PR, uses some syntax that is not supported on Node.js 6.

existentialism (approved):

PR #943, opened 2022-06-07
feat: expose webpack.target as rawTarget

bpinto:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently the target exposed via `caller.target` is the normalized target.

**What is the new behavior?**

 This change adds a new property called `rawTarget` that exposes the original target as set on the webpack configuration.

- [ ] Yes
- [x] No

* Impact:

**Other information**:

In the future, I think [browserlistEnv](https://babeljs.io/docs/en/options#browserslistenv) option could be automatically set when the webpack's [target](https://webpack.js.org/configuration/target/#browserslist) option is set using `browserslist:${env}`. I believe that would make the webpack / babel / browserslist integration easier.

I'm afraid that would be a breaking change, and since there are, based on the [documentation](https://babeljs.io/docs/en/options#no-targets), future breaking changes it might make sense to wait for that to introduce this automatic `browserlistEnv` setting.

### DIFF
--- a/src/injectCaller.js
+++ b/src/injectCaller.js
@@ -1,12 +1,13 @@
 const babel = require("@babel/core");
 
-module.exports = function injectCaller(opts, target) {
+module.exports = function injectCaller(opts, target, rawTarget) {
   if (!supportsCallerOption()) return opts;
 
   return Object.assign({}, opts, {
     caller: Object.assign(
       {
         name: "babel-loader",
+        rawTarget,
 
         // Provide plugins with insight into webpack target.
         // https://github.com/babel/babel-loader/issues/787
--- a/src/index.js
+++ b/src/index.js
@@ -161,8 +161,9 @@ async function loader(source, inputSourceMap, overrides) {
 
   // babel.loadPartialConfigAsync is available in v7.8.0+
   const { loadPartialConfigAsync = babel.loadPartialConfig } = babel;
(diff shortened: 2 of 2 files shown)


bpinto:
I have not updated any documentation nor included any test yet. But I'd like some guidance on how to do both if this change is interesting.

bpinto:
I'm not used to working on packages that require being pre-built before they can be used in a project, so if possible could you tell me how can I test this change on my project? I tried adding it to `package.json` using the git URL but that is not enough. Meanwhile, I have copied these changes over to the installed package (on `node_modules`) in order to try this out.

src/injectCaller.js
```diff
@@ -1,12 +1,13 @@
...
 
-module.exports = function injectCaller(opts, target) {
+module.exports = function injectCaller(opts, target, rawTarget) {
   if (!supportsCallerOption()) return opts;
 
   return Object.assign({}, opts, {
     caller: Object.assign(
       {
         name: "babel-loader",
+        rawTarget,
```
bpinto, line 10: Maybe I should add this below the `target` property below?

bpinto:
@nicolo-ribaudo do you think this is worth? See https://github.com/babel/babel-loader/pull/826#issuecomment-1149116915 for context. I would like to be able to use different browserslist targets.

PR #971, opened 2022-11-01
Pass external dependencies from Babel to Webpack

nicolo-ribaudo:

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

**What is the new behavior?**

We added support for `externalDependencies` in Babel 7.12.0, but we never properly wired it in `babel-loader`. `babel-loader` must forward the list of external dependencies from `@babel/core` to `webpack`, so that webpack knows which files to watch and when to re-run Babel.

- [ ] Yes
- [x] No

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -210,8 +210,9 @@ async function loader(source, inputSourceMap, overrides) {
         });
       }
 
-      const { code, map, metadata } = result;
+      const { code, map, metadata, externalDependencies } = result;
 
+      externalDependencies?.forEach(dep => this.addDependency(dep));
       metadataSubscribers.forEach(subscriber => {
         subscribe(subscriber, metadata, this);
       });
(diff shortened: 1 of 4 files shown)


nicolo-ribaudo:
I plan to backport this to v8, since we released v9 just a few days ago.

src/index.js
```diff
@@ -210,8 +210,9 @@ async function loader(source, inputSourceMap, overrides) {
         });
       }
 
-      const { code, map, metadata } = result;
+      const { code, map, metadata, externalDependencies } = result;
 
+      externalDependencies?.forEach(dep => this.addDependency(dep));
```
JLHwung, line 215: Do we use optional chaining here because we can't control the return type of `overrides.result`? If so we might consider convert `babel-loader` to ts and publish typings.

JLHwung (approved):

src/index.js
nicolo-ribaudo: Yeah, I added `?` because current `overrides.result` functions probably don't return `externalDependencies`, if they replace the result object.

Issue #983, opened 2023-02-16
Refresh time is too slow in expo react-native project after upgrade


elitenas:
**I'm submitting a bug report**

**Webpack Version:**
1.10.x/2.x

**Babel Core Version**:
7.20.12

**Babel Loader Version**:
8.3.0

**Please tell us about your environment:**
OSX 10.x / Linux / Windows 10: OSX 13.2

**Current behavior:** Watch refresh time is too slow after upgrading from 8.2.5 to 8.30 in react-native expo project

**Expected/desired behavior:** Once i make an edit on a file, the refresh should be within 5 seconds but it takes 45 secs

elitenas:
@nicolo-ribaudo

liuxingbaoyu:
Can you provide a repo to reproduce it?

elitenas:
@liuxingbaoyu thanks for the response. Here is a repro project: https://drive.google.com/file/d/1lYluHiGl4UVnp9bbPjcIOid09NnbYifX/view?usp=sharing

It takes 30 seconds for a simple text change to apply. With babel-loader 8.2.5 or below, takes 2-3 seconds max.

JLHwung:
@elitenas Thanks. I have downloaded the repro and will take a look.

JLHwung:
@elitenas Can you attach the specific reproduction steps?

Here is mine:
```
npm i
npm run web
# Manually edit src/screens/Authentication/Login.tsx
# Observe http://localhost:19006/
```

The page reload is instant, less than 3 seconds consistently.

elitenas:
@JLHwung My repro steps were similar.

1) npm install
2) npm start
3) press w
4) make a change to login page

The only thing i would point out is make sure you are on npm > 8.3.x. I am not in front of my my work laptop but I believe we are on 8.19.2. Otherwise the overrides object in package.json is ignored if you are on npm below that.

Let me know if this doesn't work. I had another developer reproduce this well. So i can double check.

JLHwung:
> Let me know if this doesn't work.
```
$ npm -v
9.4.0
$ npm ls babel-loader
my-app@1.0.0 /path/to/my-app
└─┬ @expo/webpack-config@18.0.1
  └── babel-loader@8.3.0 overridden
$ cat node_modules/babel-loader/package.json | grep version
  "version": "8.3.0",
    "preversion": "yarn run test",
```

> Expo by default doesn't use babel-loader 8.3.0 yet

[...]

PR #984, opened 2023-02-23
fix: Enable cache with `externalDependencies`

liuxingbaoyu:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Fixes: #983 

**What is the new behavior?**

- [ ] Yes
- [x] No

* Impact:

**Other information**:
I'm not sure how to test it, I manually tested it successfully in the repro of #983.
This may need to be backported to 8.x as this is a regression.

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -1,1 +1,4 @@
 yarnPath: .yarn/releases/yarn-3.2.3.cjs
+enableGlobalCache: true
+nodeLinker: node-modules
+
--- a/test/cache.test.js
+++ b/test/cache.test.js
@@ -389,3 +389,90 @@ test.cb("should allow to specify the .babelrc file", t => {
     });
   });
 });
+
+test.cb("should cache external dependencies", t => {
+  const dep = path.join(cacheDir, "externalDependency.txt");
+
+  fs.writeFileSync(dep, "123");
+
+  let counter = 0;
+
+  const config = Object.assign({}, globalConfig, {
+    entry: path.join(__dirname, "fixtures/constant.js"),
+    output: {
+      path: t.context.directory,
+    },
+    module: {
+      rules: [
+        {
+          test: /\.js$/,
+          loader: babelLoader,
+          options: {
+            babelrc: false,
+            configFile: false,
+            cacheDirectory: t.context.cacheDirectory,
+            plugins: [
+              api => {
+                api.cache.never();
+                api.addExternalDependency(dep);
+                return {
+                  visitor: {
+                    BooleanLiteral(path) {
+                      counter++;
+                      path.replaceWith(
+                        api.types.stringLiteral(fs.readFileSync(dep, "utf8")),
+                      );
+                      path.stop();
+                    },
+                  },
+                };
+              },
+            ],
+          },
+        },
+      ],
+    },
+  });
+
+  webpack(config, (err, stats) => {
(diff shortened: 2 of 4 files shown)


JLHwung (changes requested):
Thanks for the PR. I am afraid this approach is not very performant: essentially we are querying fs for external dependencies stats during every webpack re-compilation, even though most of re-run may not be trigged by external dependencies.

We can obtain a set of changed files from the webpack watch file system:
```js
this._compiler.watchFileSystem.watcher.getAggregated()
// returns { changes: Set<string>, removals: Set<string> }
```

and then we can see if there are changes in external dependencies, if so we can invalidate and purge the cache, otherwise we assume the cache is safe to be reused. In this approach all the extra works are done in memory, no extra IO is involved.

Of course `this._compiler` is a webpack specific property, the docs has warned that using `this_compiler` would "have a negative impact on your loaders compatibility." We can open an issue at the webpack repo to ask for their input.

liuxingbaoyu:
> We can open an issue at the webpack repo to ask for their input.

Maybe in webpack5 we don't need to implement our own caching.

liuxingbaoyu:
image
Unfortunately this doesn't seem to work.

JLHwung:
> > We can open an issue at the webpack repo to ask for their input.
> 
> Maybe in webpack5 we don't need to implement our own caching.

Well `babel-loader@9` does support webpack 5 only. You can explore if we can get rid of our own cache and count on the webpack persistent cache.

JLHwung:
> image Unfortunately this doesn't seem to work.

Does it work if we pass absolute file paths to `this.addDependency`?

liuxingbaoyu:
Should already be an absolute path in this test.

liuxingbaoyu:
Yes, I plan to finish the backport first to fix the regression, then migrate the code to ts, and then try to refactor the cache.

.yarnrc.yml
```diff
@@ -1 +1,4 @@
 yarnPath: .yarn/releases/yarn-3.2.3.cjs
+enableGlobalCache: true
+nodeLinker: node-modules
```
liuxingbaoyu, line 3: My `prettier-vscode` doesn't seem to be working properly.

test/cache.test.js
```diff
@@ -389,3 +389,90 @@ test.cb("should allow to specify the .babelrc file", t => {
...
+
+    t.is(counter, 1);
+
+    webpack(config, (err, stats) => {
+      t.deepEqual(stats.compilation.warnings, []);
+      t.deepEqual(stats.compilation.errors, []);
+
+      t.true(stats.compilation.fileDependencies.has(dep));
+
+      t.is(counter, 2);
```
liuxingbaoyu, line 451: Unfortunately this still caches twice.

.yarnrc.yml
JLHwung: Did you try https://yarnpkg.com/getting-started/editor-sdks?

JLHwung (approved):
Can you retrieve the filestamps of external dependencies from
```
this._compilation.fileSystemInfo.getFileTimestamp(path: string, callback)
```
webpack will add it to the builtin file timestamp cache, if cache is not hit, it will parallelize the filesystem stat call.

I realize that the watcher approach does not cover all situation since changes on external dependencies might happen out of the webpack watch cycle. So mtime is the correct approach.

PR #995, opened 2023-07-06
bump find-cache-dir to v4

JLHwung:
**Please Read the [CONTRIBUTING Guidelines](https://github.com/babel/babel-loader/blob/main/CONTRIBUTING.md)**
*In particular the portion on Commit Message Formatting*

**Please check if the PR fulfills these requirements**
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: Bump `find-cache-dir` to v4

**What is the current behavior?** (You can also link to an open issue here)

The current `find-cache-dir` depends on `make-dir` which depends on `semver@6.3.0`.

**What is the new behavior?**

Behaviour is not changed.

**Does this PR introduce a breaking change?**
- [ ] Yes
- [x] No

If this PR contains a breaking change, please describe the following...

* Impact:
* Migration path for existing applications:
* Github Issue(s) this is regarding:

**Other information**: Closes #994

PR #1005, opened 2023-08-12
Fix esm compatibility with jest

ahce:
- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Steps to reproduce:
1. Create `package.json` file
```json
{
  "scripts": {
    "test": "jest"
  },
  "devDependencies": {
    "babel-loader": "9.1.3",
    "jest": "29.6.2"
  }
}
```
2. Create `babel.test.js` file
```js
test('should import babel-loader correctly', () => {
  expect(require('babel-loader')).toBeInstanceOf(Function);
});
```
3. Run test
Test fail: `Error: You need to run with a version of node that supports ES Modules in the VM API. See https://jestjs.io/docs/ecmascript-modules`

**What is the new behavior?**

The test pass successfully
image

- [ ] Yes
- [x] No

### DIFF
--- a/src/cache.js
+++ b/src/cache.js
@@ -13,7 +13,6 @@ const zlib = require("zlib");
 const crypto = require("crypto");
 const { promisify } = require("util");
 const { readFile, writeFile, mkdir } = require("fs/promises");
-const findCacheDirP = import("find-cache-dir");
 
 const transform = require("./transform");
 // Lazily instantiated when needed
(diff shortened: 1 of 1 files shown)


nicolo-ribaudo:
Does this actually fix compatibility? You can now require it, but can you _use_ it in jest?

ahce:
> Thanks! Does this actually fix compatibility? You can now require it, but can you _use_ it in jest?

Thanks for your reply! 
I have a custom `webpack plugin` integrated with `babel-loader`, and from version `9.1.3` my tests fail.
I detected that in `v9.1.3 `this change was introduced in #995.
I applied this fix directly in `node_modules` and my tests work correctly.

PR #1026, opened 2024-07-12
bump node requirement to `^18.20.0 || ^20.10.0 || >=22.0.0` and webpack requirement to `>= 5.61.0`

JLHwung:
**Please Read the [CONTRIBUTING Guidelines](https://github.com/babel/babel-loader/blob/main/CONTRIBUTING.md)**
*In particular the portion on Commit Message Formatting*

**Please check if the PR fulfills these requirements**
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

**What is the current behavior?** (You can also link to an open issue here)
Behaviour is not expected to change.

**What is the new behavior?**

**Does this PR introduce a breaking change?**
- [x] Yes
- [ ] No

If this PR contains a breaking change, please describe the following...

* Impact: Medium
  Dropped support for dead node.js versions and old webpack 5 versions.
* Migration path for existing applications: Upgrade node.js to `^18.20.0 || ^20.10.0 || >=22.0.0` and webpack to `>= 5.61.0` or pin babel-loader to 9
* Github Issue(s) this is regarding:

**Other information**:
The webpack requirement is bumped to 5.61.0 as this is the first webpack version that supports the Node.js 18 change without the `--openssl-legacy-provider` switch: https://github.com/webpack/webpack/commit/0f6c78cca174a73184fdc0d9c9c2bd376b48557c

PR #1027, opened 2024-07-15
breaking: use output.hashFunction as loader cache

JLHwung:

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The babel-loader hash function can not be changed by users. It will use `sha256` and then fallback to `md5`.

**What is the new behavior?**
The hash function is now determined by the webpack option `output.hashFunction`. This opens the door to more performant hash such as the future webpack defaults `xxhash64`. Plus we don't have to update the hash function again if in the future OpenSSL drops more unsafe crypto hash functions, since users can provide the desired `hashFunction` in the webpack config.

- [x] Yes
- [ ] No

The cache filename might be changed because now it aligns to the webpack option `output.hashFunction`, which may be different than our previous hash function choice.

* Impact: Low
* Migration path for existing applications: purge the babel-loader cache before upgrading babel-loader. If they are not purged, the function of babel-loader will not be impacted but doing so will save some disk space.

**Other information**:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -181,6 +181,9 @@ async function loader(source, inputSourceMap, overrides) {
 
     let result;
     if (cacheDirectory) {
+      const hash = this.utils.createHash(
+        this._compilation.outputOptions.hashFunction,
+      );
       result = await cache({
         source,
         options,
@@ -188,6 +191,7 @@ async function loader(source, inputSourceMap, overrides) {
         cacheDirectory,
         cacheIdentifier,
         cacheCompression,
+        hash,
       });
     } else {
       result = await transform(source, options);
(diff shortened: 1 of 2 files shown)


JLHwung:
Rebased on #1026 as the `utils.createHash` API is not available on webpack 5.0.0.

src/index.js
```diff
@@ -181,13 +181,17 @@ async function loader(source, inputSourceMap, overrides) {
 
     let result;
     if (cacheDirectory) {
+      const hash = this.utils.createHash(
+        this._compilation.outputOptions.hashFunction,
+      );
```
JLHwung, line 186: Usage from https://github.com/webpack/webpack/blob/277460b33bcc49c51acbbcd688672aa4ec685732/lib/DefinePlugin.js#L361

nicolo-ribaudo (approved):

liuxingbaoyu (approved):

PR #1032, opened 2024-08-22
Create SECURITY.md

JLHwung:
Drafted security policy for `babel-loader`.


nicolo-ribaudo (approved):
I think it's fine to only support the latest major, as weback 5 was released in 2020.

PR #1033, opened 2024-08-26
Support cache with external dependencies

### LINKED ISSUE #983: Refresh time is too slow in expo react-native project after upgrade

JLHwung:

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The babel-loader cache is disabled when there are external dependencies registered from the Babel plugin API `addExternalDependencies`

**What is the new behavior?**
Still enable the cache if there are external deps. When the external deps are modified, the cache will be disgarded.

- [ ] Yes
- [x] No

* Impact:

**Other information**:
This PR is reviving #984, it picks the tests from #984 but adopts a slightly different approach.

In #984 we generate timestamps and use that as a cache identifier. As is pointed out in https://github.com/babel/babel-loader/pull/984#discussion_r1146670461, the cache will not be reused after the first run because of different cache identifiers (One for unknown externalDeps and the other for known externalDepsMtimes).

In this PR we store the timestamps of external deps within the transform result and validate it against the new timestamps returned from the webpack API `_compilation.fileSystemInfo.getFileTimestamp`, so the first cache will be reused because there is no cache identifier changes as now we determine if the cache should be reused _after_ reading the cache result.

Closes #984, fixes #983.

### DIFF
--- a/src/cache.js
+++ b/src/cache.js
@@ -65,10 +65,7 @@ const filename = function (source, identifier, options, hash) {
   return hash.digest("hex") + ".json";
 };
 
-const handleExternalDepedencies = async function (
-  externalDependencies,
-  getFileTimestamp,
-) {
+const addTimestamps = async function (externalDependencies, getFileTimestamp) {
   for (const depAndEmptyTimestamp of externalDependencies) {
     try {
       const [dep] = depAndEmptyTimestamp;
@@ -78,7 +75,6 @@ const handleExternalDepedencies = async function (
       // ignore errors if timestamp is not available
     }
   }
-  return externalDependencies;
 };
 
 const areExternalDependenciesModified = async function (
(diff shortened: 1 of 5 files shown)


src/cache.js
```diff
@@ -65,6 +65,42 @@ const filename = function (source, identifier, options, hash) {
...
 
+const handleExternalDepedencies = async function (
+  externalDependencies,
+  getFileTimestamp,
+) {
+  const result = [];
+  for (const dep of externalDependencies) {
+    try {
+      const { timestamp } = await getFileTimestamp(dep);
+      result.push(dep + "\0" + timestamp);
```
liuxingbaoyu, line 76: Can I use [dep, timestamp] here and sort externalDependencies in advance?

liuxingbaoyu (approved):
Sorry I forgot about that one. Looks great, thanks!

PR #932, opened 2022-03-06
Use LMDB cache instead of regular FS cache to speed up cache hits

pastelsky:
**Please check if the PR fulfills these requirements**
- [x] Tests for the changes have been added (for bug fixes / features)
- [] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe:

Improved performance of babel-loader when files are cached. LMDB is much faster to access than the regular file cache used by babel-loader

**What is the current behavior?** (You can also link to an open issue here)
`babel-loader` saves files for caching built contents.

**What is the new behavior?**
Replaces the use of individual files with [lmdb](https://www.npmjs.com/package/lmdb), which improved babel-loader times, especially when hitting caches in large projects.
LMDB is already used by projects such as  [Parcel](https://parceljs.org/) and [Elasticsearch's Kibana](https://www.elastic.co/kibana/), as the storage layer for [HarperDB](https://harperdb.io/) and [Gatsby](https://www.gatsbyjs.com/) for fast memory-mapped cached access.

This has a side-effect of making babel-loader's implementation a bit simpler as well

**Does this PR introduce a breaking change?**
- [ ] Yes
- [n] No

PR #1040, opened 2024-09-04
breaking: use webpack builtin cache

### LINKED ISSUE #558: Use hashed directory structure

**Current behavior:**

Currently all outputs are stored into a single cache directory. On large projects and certain hardware this might have a negative impact on performances.

Something that might remedy the situation is to write into subfolders that contain the first two characters of the key used for storing and use these for reading the cached output.

### LINKED ISSUE #634: Prune babel cache?

### LINKED ISSUE #816: Can u provide write and read options to override default behavior
**Webpack Version:**
4

**Babel Core Version**:
7

**Babel Loader Version**:
8

Macos

**What is the expected behavior?**

Just like [cache-loader](https://github.com/webpack-contrib/cache-loader#options) that provide `read` and `write` options. I can cache my data to redis or memcache even memory to make webpack rebuild faster.

### LINKED ISSUE #831: Expose cache implementation

Is there any possibility to expose the [cache implementation](https://github.com/babel/babel-loader/blob/master/src/cache.js) as a direct export? Or as a separate package?

A form of this is already done in https://github.com/viankakrisna/loader-fs-cache but already seems to be somewhat out of sync with this version.
It's used in [eslint-loader](https://github.com/webpack-contrib/eslint-loader/blob/master/package.json#L45) as their caching mechanism.

Pulling it directly from here seems like a cleaner solution and I imagine there are multiple other packages that would like to use it. 

I realize that there's a possibility to import it like `import cache from 'babel-loader/lib/cache'` but that seems iffy and could break easily.

### LINKED ISSUE #978: Cache files take up too much disk space

**Webpack Version:**
5.74.0

**Babel Core Version**:
7.18.10

**Babel Loader Version**:
8.2.5

OSX 13.1

**Current behavior:**

For the cache directory, a cache file will be generated every time the code is changed, which will cause a lot of unnecessary disk usage. I found that some of my projects have more than 1G cache usage.

**Expected behavior:**

It is expected that one source file corresponds to one cache file. Every time a file is changed, only the old cached file is overwritten, no new one is added. thanks.

JLHwung:

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: Use webpack builtin cache

The babel-loader uses per-input filesystem cache.

**What is the new behavior?**
The babel-loader option `cacheCompression` will be removed, users can configure the webpack option `cache.
babel-loader will use the builtin webpack cache.

The builtin webpack cache offers many options and users will be able to compose their own cache plugin by tapping the compiler cache hooks.

- [x] Yes
- [ ] No

* Impact: Big
If you are using the babel-loader `cacheDirectory` or `cacheCompression` option, set `cacheDirectory` to `true` and enable the webpack persistent cache via
```js
// webpack.config.js
{
  cache: {
    type: "filesystem",
    compression: "gzip"
  }
}
```
The default webpack fs cache location is `./node_modules/.cache/webpack`. You can change the cache directory via the webpack option `cache.cacheDirectory`. See the [official docs](https://webpack.js.org/configuration/cache/) for more webpack cache options.

Closes #932, as LMDB cache can be implemented via a third-party webpack cache plugin
Closes #978 and closes #634, as you can instruct webpack to store unused filesystem cache up to one-month via `cache.maxAge`
Closes #831, as you can implement your own webpack cache plugin
Closes #816, as redis cache can be implemented via a third-party webpack cache plugin
Closes #558, as webpack save multiple files into a single pack when the cache type is filesystem

**Other information**:

This is an early WIP, obviously most cache tests are currently failing due to config errors and incorrect filesystem structures. The updated cache test serves as a proof-of-concept that this approach works with the memory cache (The default of webpack 5 in development mode). Note that the memory cache will be purged when the compiler shutdown, so for those who enabled the babel-loader cache, they should enable the webpack filesystem cache.

When the filesystem cache is enabled, users can simply disable the Babel cache in most circumstances, such as 1) babel-loader is the only webpack loader for the js/ts inputs or 2) every js/ts loader is cachebale (the webpack defaults for all loaders unless the loader explicitly called `this.cacheable(false)`). The transformed output will be cached by webpack into the filesystem even without babel-loader cache. So in the proof-of-concept test we have to plug in an uncacheable loader for babel-loader cache to be hit again.

**Alternative Solutions**:

A `cacheProvider: "webpack" | "babel-loader" ` option to switch between webpack cache and our own cache. But the new option will further complicates the cache config, and the `babel-loader` cache does not offer very much functionalities than the webpack builtin cache.

I would like to get feedback from team members, @liuxingbaoyu and @nicolo-ribaudo Do you think this approach is a good direction or we should keep our own cache? Do you think we should ship this in v10 or defer the changes to v11?


src/index.js
```diff
@@ -20,13 +20,12 @@ if (/^6\./.test(babel.version)) {
 }
 
 const { version } = require("../package.json");
-const cache = require("./cache");
```
JLHwung, line 23: The `./cache` is now unused, we can remove this file and the only dependency `find-up` after tests are fixed.

src/cacheHandler.js
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+
+  if (!result) {
+    logger.debug("applying Babel transform");
+    result = await transform(source, options);
+    await addTimestamps(result.externalDependencies, getFileTimestamp);
+    logger.debug(`caching result for '${filename}'`);
+    await itemCache.storePromise(result);
+    logger.debug(`cached result for '${filename}'`);
```
JLHwung, line 95: The cache log is updated from "writing result to cache file" to current wordings to reflect the fact that the loader is unaware of the cache implementation: it may be written to the filesystem, it may be written to a database or it is retained in the memory.

liuxingbaoyu:
> Do you think this approach is a good direction or we should keep our own cache? Do you think we should ship this in v10 or defer the changes to v11?

This looks good to me. Both v10 and v11 are ok for me and the migration steps seem to be simple.

JLHwung:
This PR is ready for review. The options docs are also updated.

liuxingbaoyu:
Can we still utilize webpack cache system cache when `webpack` options disable cache?🤔

JLHwung:
> Can we still utilize webpack cache system when `webpack` options disable cache?🤔

In most case, no. When webpack cache is disabled, the builtin cache is doing nothing. The cache facade provided by webpack is essentially an event emitter. In this case, the cache action event is triggered but there is no cache event handler to actually do the cache job.

If this PR gets merged, I would also note at the release note that to enable the babel-loader cache, you will have to turn on the webpack cache. But for most cases, the webpack cache is enough and you don't have to enable the babel-loader cache.

steverep:
Will this break compatibility with Rspack?

nicolo-ribaudo (approved):
Nice

JLHwung:
Let's postpone this PR until https://github.com/babel/babel/issues/16915#issuecomment-2423368393 is addressed.

JLHwung:
> Will this break compatibility with Rspack?

Good question. We don't have integration tests with Rspack so I am not sure if babel-loader can work with Rspack. In this PR we use the webpack builtin cache API, if Rspack support that API, as it said it is "webpack compatible", then babel-loader in this PR should work with Rspack.

PR #1053, opened 2025-05-06
perf: switch to `empathic` find package traversal

43081j:
Uses `empathic/find` for finding the closest `package.json`, a much faster and smaller library than the various over-specialised alternatives lying around.

- [ ] Bugfix
- [ ] Feature
- [ ] Docs update
- [ ] Build related changes
- [ ] CI related changes

We traverse the file system for the closest package.json-containing directory.

**What is the new behavior?**

N/A

- [ ] Yes
- [x] No


43081j:
Over time, the de-dupe will switch since we're in the process of moving many other popular tools to do this switch too.

you're right that the performance bump won't be much here since it isn't called often, but it is greater than none. there's currently no reason to not do this switch other than doubt that the de-dupe will be improving, but we have to start somewhere.

scutuatua-crypto (approved):

